PR #24, opened 2022-03-08
Add jupyterlite_contents config

martinRenou:


docs/conf.py
```diff
@@ -8,6 +8,7 @@
 html_logo = "_static/icon.svg"
 
 jupyterlite_config = "jupyterlite_config.json"
+jupyterlite_contents = "./custom_contents"
```
jasongrout, line 11: I thought this was a list, not a string?

martinRenou: Good catch! Thanks

martinRenou: Yes, that makes sense. I'm pushing a last commit to allow this.

src/jupyterlite_sphinx.py
```diff
@@ -267,7 +270,25 @@ def jupyterlite_build(app: Sphinx, error):
         if app.env.config.jupyterlite_config:
             config = ["--config", app.env.config.jupyterlite_config]
 
+        contents = []
+        if app.env.config.jupyterlite_contents:
+            for content in app.env.config.jupyterlite_contents:
+                contents.extend(["--contents", content])
```
jasongrout:
```suggestion
        contents = []
        dirs = app.env.config.jupyterlite_contents
        if dirs:
            if isinstance(dirs, str):
                dirs = [dirs]
            for content in dirs:
                contents.extend(["--contents", content])

```
This makes it possible for the value to be a single directory, which I bet is a very common case.

src/jupyterlite_sphinx.py
```diff
@@ -267,7 +270,31 @@ def jupyterlite_build(app: Sphinx, error):
...
             config = ["--config", app.env.config.jupyterlite_config]
 
+        contents = []
+        if app.env.config.jupyterlite_contents:
+            jupyterlite_contents = app.env.config.jupyterlite_contents
+
+            if isinstance(contents, str):
+                contents.extend(["--contents", jupyterlite_contents])
+
+            if isinstance(contents, (tuple, list)):
```
jasongrout, line 280:
```suggestion
            if isinstance(jupyterlite_contents, str):
                contents.extend(["--contents", jupyterlite_contents])

            if isinstance(jupyterlite_contents, (tuple, list)):
```

martinRenou: 😅

PR #52, opened 2022-06-16
Allow whitespace in filenames

jasongrout:
Without this, a space in a filename makes it into two separate arguments.

This aligns with the literalinclude directive, which also takes filenames:

https://github.com/sphinx-doc/sphinx/blob/7bdc11e87c7d86dcc2a087eccb7a7c129a473415/sphinx/directives/code.py#L374


martinRenou (approved):
Thanks!

PR #54, opened 2022-06-16
Adopt the path convention of other directives like literalinclude

jasongrout:
We adopt the conventions of the sphinx [literalinclude](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) directive using [similar logic](https://github.com/sphinx-doc/sphinx/blob/7bdc11e87c7d86dcc2a087eccb7a7c129a473415/sphinx/directives/code.py#L411-L412):

- If the path is absolute, it is considered as relative to the document root (after stripping off the initial /)
- If the path is relative, it is considered as relative to the source document


martinRenou (approved):
LGTM, thanks a lot!

PR #58, opened 2022-06-23
More consistent naming

### LINKED ISSUE #57: Default config file option is inconsistent with jupyterlite
It's a little confusing that the config file name option here is `jupyterlite_config` (and in ipyleaflet, the config filename is `jupyterlite_config.json`), while in jupyterlite the default name is `jupyter_lite_config.json`.

martinRenou:
Fix #57

cc. @jasongrout


src/jupyterlite_sphinx.py
```diff
@@ -266,19 +266,33 @@ def jupyterlite_build(app: Sphinx, error):
...
+            jupyter_lite_config = app.env.config.jupyterlite_config
 
-        contents = []
-        if app.env.config.jupyterlite_contents:
-            jupyterlite_contents = app.env.config.jupyterlite_contents
+        if app.env.config.jupyterlite_dir:
+            warn(
+                "jupyterlite_dir config option is deprecated, please use jupyter_lite_dir instead"
+            )
+            jupyter_lite_dir = app.env.config.jupyterlite_dir
```
martinRenou, line 284: Note that we don't need a deprecation warning for the `jupyterlite_contents` rename because it has not been released yet

jasongrout:
+1, especially while we're still pretty small and growing.

Just double checking - is it consistently `jupyter_lite` (i.e., two words) over in JupyterLite for all the various user APIs, not just for jupyter_lite_conf

jasongrout (approved):

jasongrout:
I can't merge, but I can approve. Thanks!

martinRenou:
You should have received an invitation to the jupyterlite org and for having commit rights on this repo :) You probably need to accept it. I let you merge then! :)

jasongrout:
On second thought, I looked through the jupyterlite repo, and they are more consistently treating jupyterlite as one word in the code, even if the config file naming is treating jupyterlite as two words. I think we ought to hold off on the breaking change here to get more clarity on which way we should consistently name things. It may be that they decide to change jupyterlite to have the config file be named `jupyterlite_config.json`?

jtpio:
Probably this is unlikely to change in jupyterlite, and I believe `jupyter_lite_config.json` was chosen since it's similar to `jupyter_notebook_config.json` or `jupyter_server_config.json`.

jasongrout:
> Probably this is unlikely to change in jupyterlite, and I believe `jupyter_lite_config.json` was chosen since it's similar to `jupyter_notebook_config.json` or `jupyter_server_config.json`.

Thanks for chiming in, and for some of the context.

martinRenou:
Arf sorry I merged so fast then :S

I agree with you, let's revert and release again.

jasongrout:
> I agree with you, let's revert and release again.

+1 - I think this would be 0.6.0, since this is technically breaking the 0.5.0 api. I don't think we need to put in backwards compatible shims, though, on the thought that most people would not have upgraded to 0.5.0.

PR #59, opened 2022-06-23
Update Pypi description to be the same as the readme

jasongrout:
Right now pypi says there is no description, so we just use the standard way to read in the readme file.


martinRenou (approved):
Great thanks!

PR #60, opened 2022-06-28
Remove unneeded code

jtpio:
Looking at the source, it seems like this snippet is not needed anymore.


martinRenou (approved):
Thanks! This might be a leftover during a rebase with conflicts

PR #64, opened 2022-07-16
add globbing to content

amueller:
This would be useful for integrating with sphinx-gallery (maybe).
It allows setting the content to all the generated notebooks quite easily.
The other option would be to take the generated notebooks and move them into their own folder and then add that folder.


martinRenou (approved):
Thanks!

PR #67, opened 2022-07-25
"Try It Live!" button

### LINKED ISSUE #50: Add a "Try it live!" button that the user needs to press prior to showing the embedded iframe
cc. @davidbrochart

### Problem

When jupyterlite-sphinx embeds JupyterLite in a documentation page, it can be tens of MB for the viewer to download, this can be really annoying if the viewer is not interested in testing JupyterLite but just want to see the rest of the docs.

### Proposed Solution

We should add the possibility to show a button "Try it live!" that the viewer needs to press before the iframe is added to the page.

Wishlist for this feature:
- make this an opt-in or an opt-out
- make the "Try it live!" prompt customizable

martinRenou:
Fix #50

tryitbutton

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,4 +37,4 @@ jobs:
         channels: conda-forge
 
     - name: Test PEP8
-      run: black --check src
+      run: black --check jupyterlite_sphinx
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-src/jupyterlite_sphinx.egg-info
+jupyterlite_sphinx/jupyterlite_sphinx.egg-info
 **/.ipynb_checkpoints
 **/__pycache__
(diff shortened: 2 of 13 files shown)


jtpio:
Looking good!

Wondering whether adding the "thread" of the bulb would make it even more explicit this is an embedded JupyterLite app? 

image

martinRenou:
Yes, I was also thinking of mimicking slightly the JupyterLab spinner with 3 yellow moons rotating around the bulb. What do you think?

I implemented this using pure CSS, so no SVG involved. Maybe using SVG could simplify the setup?

martinRenou:
@jtpio if you don't mind, we could go ahead with the current state of this PR and discuss look improvements ideas in an issue?

jtpio:
Sure thing.

PR #72, opened 2022-08-16
Make the jupyterlite_contents glob recursive.

jasongrout:
Since we have shifted to using globs for the contents, we might as well enable the strictly more powerful ** recursive syntax. See https://docs.python.org/3/library/glob.html#glob.glob

"If recursive is true, the pattern “**” will match any files and zero or more directories, subdirectories and symbolic links to directories. If the pattern is followed by an os.sep or os.altsep then files will not match."


martinRenou (approved):
Thank you!

PR #73, opened 2022-08-17
Restore jupyterlite_contents being optionally a string

jasongrout-db:
Restore jupyterlite_contents being a string (which was lost when we moved to using globs). Also, add some documentation about values for jupyterlite_contents.


martinRenou (approved):
Thank you!

PR #62, opened 2022-06-28
Revert renaming the config properties

martinRenou:
cc. @jasongrout @jtpio

PR #74, opened 2022-08-17
Finish reverting config names.

jasongrout:
In #62, we reverted the config name changes from #58. However, we forgot to revert the changes to our docs conf.py config as well.


martinRenou (approved):
Thanks!

PR #78, opened 2022-11-30
DOC fix broken link to custom Jupyterlite configuration

lesteve:
The link is currently broken. Not sure whether there is a more appropriate link that the one I used and ends up in the middle of the HowTo guide.

To get an idea about the original content the link was pointing to, see:
https://github.com/jupyterlite/jupyterlite/blob/dc395c42d610a0bc6f96bb715a3f130e2ef7c43a/docs/configuring.md


martinRenou (approved):
Thanks!

PR #85, opened 2023-02-03
Unpin sphinx

### LINKED ISSUE #83: release the pin on Sphinx version ?
In the current implementation, jupyterlite-shinx enforces a Sphinx 4 version. Sphinx 6 is out since December and Sphinx 4 is starting to be dropped by other Sphinx extensions. 

Would it be possible to change/remove this pinned version from setup.py ?

lesteve:
fix #83
close #84

### DIFF
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -182,7 +182,6 @@ class RepliteDirective(SphinxDirective):
 
 
 class _LiteDirective(SphinxDirective):
-
     has_content = False
     optional_arguments = 1
     final_argument_whitespace = True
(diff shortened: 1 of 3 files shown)


lesteve:
Trying to work-around the conda conflict issue in the CI which is likely related to https://github.com/conda-incubator/setup-miniconda/issues/274

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -182,7 +182,6 @@ def run(self):
 
 
 class _LiteDirective(SphinxDirective):
-
```
lesteve, line 185: Somehow latest black says there should be no newline here.

martinRenou (approved):
LGTM!

PR #89, opened 2023-03-09
Depend on `jupyterlite-core`

jtpio:
Similar to https://github.com/jupyterlite/xeus-python-kernel/pull/114

The core functionality of the `jupyterlite` package is being extracted to a new `jupyterlite-core` package in https://github.com/jupyterlite/jupyterlite/pull/994.

This will allow site deployers to make slimmer deployments without having to bring the Pyodide kernel by default.

### TODO

- [x] Try depending on `jupyterlite-core` built from https://github.com/jupyterlite/jupyterlite/pull/994
- [x] Update to the real `jupyterlite-core` dependency when a release is available (likely after https://github.com/jupyterlite/jupyterlite/pull/994 is merged)
- [ ] Update the docs environment
- [x] Re-add a Python kernel to the docs environment

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,9 @@ dependencies = [
     "sphinx>=4",
 ]
 
+[tool.hatch.metadata]
+allow-direct-references = true
+
 [tool.hatch.version]
 path = "jupyterlite_sphinx/__init__.py"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dependencies = [
     "docutils",
     "jupyter_server",
     "jupyterlab_server",
-    "jupyterlite-core[piplite] >=0.1.0b19",
+    "jupyterlite-core >=0.1.0b19",
     "sphinx>=4",
 ]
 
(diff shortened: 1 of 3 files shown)


jtpio:
Docs seem to be building fine:

image

We can add the Pyodide as `federated_extension` to get the previous behavior of having the Pyodide kernel available by default on the docs. We could also choose Xeus Python.

jtpio:
Looking good now after updating to `jupyterlite-core>=0.1.0b19`: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.1.0b19

image

pyproject.toml
```diff
@@ -16,10 +16,13 @@ dependencies = [
     "docutils",
     "jupyter_server",
     "jupyterlab_server",
-    "jupyterlite[piplite]",
+    "jupyterlite-core[piplite] >=0.1.0b19",
```
martinRenou, line 19: is `[piplite]` still required?

jtpio: ah good catch, thanks!

martinRenou (approved):
Thanks!

PR #91, opened 2023-03-15
Add notice about `jupyterlite-core` in the changelog

jtpio:
Follow-up to the `0.8.0` release: https://github.com/jupyterlite/jupyterlite-sphinx/releases/tag/v0.8.0

This will make it more visible what the switch to `jupyterlite-core` means.


martinRenou (approved):
Thanks!

PR #94, opened 2023-03-16
Update docs to mention adding other kernels

jtpio:
Follow-up to #89 and the `0.8.0` release.

- [x] Update the installation page to mention how to install a Python kernel
- [x] Update the configuration page

image

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -18,15 +18,15 @@ extensions = [
 
 JupyterLite should automatically show up in your built online documentation. To preview it locally, you can navigate to the build directory (e.g. `_build/html`) and use `python -m http.server` to serve the site.
 
-```{note}
+````{note}
 By default `jupyterlite-sphinx` does not install a Python kernel.
 If you would like have a Python kernel available in your docs you can install either `jupyterlite-pyodide-kernel` or `jupyterlite-xeus-python` with `pip`:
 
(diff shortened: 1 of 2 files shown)


docs/installation.md
```diff
@@ -17,3 +17,16 @@ extensions = [
 ```
 
 JupyterLite should automatically show up in your built online documentation. To preview it locally, you can navigate to the build directory (e.g. `_build/html`) and use `python -m http.server` to serve the site.
+
+```{note}
+By default `jupyterlite-sphinx` does not install a Python kernel.
+If you would like have a Python kernel available in your does you can install either `jupyterlite-pyodide-kernel` or `jupyterlite-xeus-python` with `pip`:
```
martinRenou, line 23:
```suggestion
If you would like have a Python kernel available in your docs you can install either `jupyterlite-pyodide-kernel` or `jupyterlite-xeus-python` with `pip`:
```

martinRenou (commented):
Not sure if it's a Github bug, but it renders a bit weirdly in the Github files preview:

Screenshot from 2023-03-16 11-08-34

jtpio:
It's probably because of the nested code block in a note block.

But it seems to be fine on RTD.

martinRenou (approved):
Looks great!

PR #96, opened 2023-04-05
Update to `jupyterlite-core==0.1.0`, require Python 3.8

jtpio:
Couple of maintenance tasks:

- Update to the latest release: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.1.0
- Require Python 3.8
- Streamline the ReadTheDocs preview

### DIFF
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -5,9 +5,8 @@ dependencies:
     - pip
     - jupyter_server
     - jupyterlab_server
+    - jupyterlite-core >=0.1.0,<0.2
     - pydata-sphinx-theme
     - docutils
     - sphinx
     - black
-    - pip:
-        - jupyterlite-core==0.1.0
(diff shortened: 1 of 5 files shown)


jtpio:
A follow-up to this would be to make a new release, and then package `jupyterlite-sphinx` on conda forge.

dev-environment.yml
```diff
@@ -10,5 +10,4 @@ dependencies:
     - sphinx
     - black
     - pip:
-        - jupyterlite-core==0.1.0b19
-
+        - jupyterlite-core==0.1.0
```
martinRenou, line 13: Maybe we could install `jupyterlite-core` with mamba instead of pip?

martinRenou (commented):
Thanks!

martinRenou (approved):
Thanks!

PR #97, opened 2023-04-14
Add conda instructions and docs scripts

jtpio:
Update the docs to add instruction for installing `jupyterlite-sphinx` with `conda` and `mamba`.

Also add scripts to build the docs.


jtpio:
Forgot we had that one still opened. Looking good on RTD:

image

PR #98, opened 2023-05-09
Use xeus-python in docs

martinRenou:


martinRenou:
It's working, but pyolite is faster to start by a good factor. We should have jupyterlite-xeus-python work against the latest empack version in order to get better perf.

martinRenou:
Latest xeus-python is way faster :rocket: and packages would be properly cached by the browser. Merging!

Issue #107, opened 2023-09-07
Transfert search parameters from Page URL to Jupyterlite URL


brichet:
I wonder if we could transfert some search parameters from the URL of the page to the `iframeSrc` options.

### Problem

It is currently not possible to retrieve the search parameters of the page in the Jupyterlite application.

### Proposed Solution

Add a new option to the jupyterlite directive with a list of parameters to transfert to the `iframeSrc` (if the parameters exist).

### Additional context

This could be a solution for https://github.com/jupyterlite/litegitpuller/issues/6, because `litegitpuller` currently use the search param of the jupyterlite URL.

PR #108, opened 2023-09-07
Transfer search parameters from page URL to jupyterlite

brichet:
This PR adds a new directive to allow transfering some search params from the documentation URL to the jupyterlite URL.

Fixes https://github.com/jupyterlite/jupyterlite-sphinx/issues/107

### DIFF
--- a/docs/directives/jupyterlite.md
+++ b/docs/directives/jupyterlite.md
@@ -36,7 +36,7 @@ You can also pass a Notebook file to open automatically:
    :prompt_color: #00aa42
 ```
 
-
 The directive `search_params` allows to transfer some search parameters from the documentation URL to the Jupyterlite URL.\
 Jupyterlite is than able to these parameters from its own URL.\
 For example `:search_params: param1, param2` will transfer the parameters *param1* and *param2*.
+Transferring all the parameters: `:search_params: =all`
--- a/docs/directives/jupyterlite.md
+++ b/docs/directives/jupyterlite.md
@@ -35,3 +35,8 @@ You can also pass a Notebook file to open automatically:
    :prompt: Try JupyterLite!
    :prompt_color: #00aa42
 ```
+
+
+The directive `search_params` allows to transfer some search parameters from the documentation URL to the Jupyterlite URL.\
+Jupyterlite is than able to these parameters from its own URL.\
+For example `:search_params: param1, param2` will transfer the parameters *param1* and *param2*.
--- a/docs/directives/jupyterlite.md
+++ b/docs/directives/jupyterlite.md
@@ -37,6 +37,6 @@ You can also pass a Notebook file to open automatically:
 ```
 
 The directive `search_params` allows to transfer some search parameters from the documentation URL to the Jupyterlite URL.\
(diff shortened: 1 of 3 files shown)


martinRenou:
Neat! Could you add an example of this in the docs?

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -242,6 +257,10 @@ def run(self):
         prompt = self.options.pop("prompt", False)
         prompt_color = self.options.pop("prompt_color", None)
 
+        search_params = list(
```
martinRenou, line 260: Is the `search_param` directive here for whitelisting the parameters that we want to transfer? Would there be an issue not transferring all of them?

brichet: Not sure to understand.
It'ss here to build a python list from the string, but yes we could probably transfer the whole string and directly build a javascript array from it.
Do you think of an other way ?

martinRenou: I am not sure it makes sense to forward all query parameters. But it would simplify the code and user API.

brichet: OK thanks, I understand your first comment. Yes, the idea is to whitelist the parameters.

> I am not sure it makes sense to forward all query parameters. But it would simplify the code and user API.

And maybe it's better to have the option (at least with a flag) to allow the transfer or not.

brichet: Maybe we could add an option to transfer all the parameters(`=all` for example, since `=` can probably not be used as parameter name), or use only a boolean value to transfer all or none.

Do you have an opinion on it @martinRenou ?

docs/directives/jupyterlite.md
```diff
@@ -35,3 +35,8 @@ You can also pass a Notebook file to open automatically:
    :prompt: Try JupyterLite!
    :prompt_color: #00aa42
 ```
+
+The directive `search_params` allows to transfer some search parameters from the documentation URL to the Jupyterlite URL.\
+Jupyterlite will then be able to fetch these parameters from its own URL.\
+For example `:search_params: ["param1", "param2"]` will transfer the parameters *param1* and *param2*.
```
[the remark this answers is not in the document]
brichet: Why not, but I don't really know how it can be displayed to user.

The simplest solution I can find is to use something like
```
%%javascript
alert(window.location.search);
```
in a notebook, formatting a bit better the attributes.

Otherwise I didn't find a way to share value between javascript and python in a notebook, using xeus kernel.
It would be better to display the values in an output rather than using an `alert`.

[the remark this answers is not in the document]
martinRenou: Ok, let's merge it as is for now!

martinRenou:
@brichet let's get this in! Unless you'd like to push something else?

martinRenou (approved):
Thanks!

brichet:
Thanks @martinRenou for review

Issue #105, opened 2023-08-17
Generate notebooks from Examples section in docstring


steppi:
### Problem

I've been looking into ways to add interactivity to SciPy's Sphinx documentation. I've been toying around with jupyterlite-sphinx, and while it works well and gets us most of the way there, we already have examples in the documentation and it would be nice not to have to maintain separate notebooks for them. It would be great if there was a way to take the code from the Examples section of a docstring and have it converted to a notebook at docs build time. There could then be a button to change the rendered examples in-place to a corresponding JupyterLite notebook.

### Proposed Solution

Add a new directive which offers this feature. I've created a prototype and am curious if there would be interest in including this directive in jupyterlite-sphinx after it becomes more polished, or if you think it would be best if I maintain it as a separate extension?

### Additional context

The repo for the prototype is here, https://github.com/steppi/sphinx-try-examples.

You can see it deployed for a toy example here, https://steppi.github.io/interactive_doc_experiments/reference/index.html. 

For completeness, here's the repo this is deployed from https://github.com/steppi/interactive_doc_experiments.

I've created a post on the Scientific Python discourse [here](https://discuss.scientific-python.org/t/making-docstring-examples-interactive/812) if you'd like to weigh in.

I understand this is very rough, I don't really have any web development experience, but as things move forward I plan to bring in a web developer to help polish things.

Here's a screenshot.

![image](https://github.com/jupyterlite/jupyterlite-sphinx/assets/1953382/9e0f6b33-e178-47c4-8472-6306aa652551)

steppi:
> I personally think a PR would be very welcome for such a feature. Happy to provide help if needed.

Thanks @martinRenou. I especially appreciate the offer to the help. I can submit a draft PR within a week or two and we can go from there.

PR #111, opened 2023-10-20
Add try_examples directive for adding interactivity to sphinx Examples sections

steppi:
#### Reference issue
closes gh-105.

#### What does this implement/fix?
This PR adds the `try_examples` directive, which takes Examples sections like below

image

and generates notebooks based on them, providing a button to swap the example content in place with a notebook

image

#### Additional information
You can find this deployed for the SciPy documentation here, https://steppi.github.io/scipy/. I've managed to fix a lot of small bugs in notebook generation and this has finally got to the state where I haven't found any issues on that front in the SciPy docs.

For content under the `.. try_examples::` directive, doctest blocks are turned into code cells and RST text blocks are turned into markdown cells. Both inline and block math is handled so that the math displays correctly in the notebooks, as seen above.

To avoid having to manually add the directive, I've created a `global_enable_try_examples` config option, which automatically inserts the directive within Examples sections during the `autodoc-process-docstring` phase. This requires the docstring to be in the format produced by `numpydoc` during this phase ( I think this is format is also produced by `sphinx.ext.napoleon`.) It will need to be clearly documented how to set things up so that `global_enable_try_examples` will work. At the moment I've only tested this with SciPy's documentation, which has NumPy style docstrings and used the following extensions

```python
extensions = [
    'sphinx.ext.autodoc',
    'sphinx.ext.autosummary',
    'sphinx.ext.coverage',
    'sphinx.ext.mathjax',
    'sphinx.ext.intersphinx',
    'numpydoc',
    'sphinx_design',
    'scipyoptdoc',
    'doi_role',
    'matplotlib.sphinxext.plot_directive',
    'myst_nb',
    'jupyterlite_sphinx',
]
```

I've added handling for math with `sphinx.ext.mathjax`, worked around cruft added by `matplotlib.sphinxext.plot_directive` that shouldn't go into notebooks, and cleaned up the raw output for references produced by `numpydoc` so it looks nicer in the notebooks. Other extensions and workflows may cause issues, particularly if using `global_enable_try_examples`. Using the directive directly on a vanilla docstring should work fine.

For, now I've set the heights of the iframes for the generated notebooks to be the same as the heights of the rendered Examples content. This works pretty well in most cases, but looks bad for particularly short Examples sections. I plan to make this more configurable, but would appreciate feedback on the best way to do that. For now there are configuration options for the whether or not the toolbar should appear by default, the jupyterlite theme, and the css for relevant buttons. I think configuration should be more flexible.

I haven't documented the extension yet, but clear documentation will be very important.

@martinRenou I have one important question. At least for Firefox, I've found that my browser prompts me to select a kernel every time I open one of the generated notebooks, even though Pyodide is the only kernel installed. You can try that out [here](https://steppi.github.io/scipy/reference/index.html). I think this will be too annoying for users, so hope there is a way the kernel can be specified in advance.

### DIFF
--- a/jupyterlite_sphinx/jupyterlite_sphinx.js
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.js
@@ -51,17 +51,16 @@ window.tryExamplesShowIframe = (
     let iframe = iframeContainer.querySelector('iframe.jupyterlite_sphinx_raw_iframe');
 
     if (!iframe) {
-	const examples = examplesContainer.querySelector('.try_examples_content');
-	iframe = document.createElement('iframe');
-	iframe.src = iframeSrc;
-	iframe.style.width = '100%';
-	iframe.style.height = `${examples.offsetHeight}px`;
-	iframe.classList.add('jupyterlite_sphinx_raw_iframe');
-	examplesContainer.classList.add("hidden");
-	iframeContainer.appendChild(iframe);
-    }
-    else {
-	examplesContainer.classList.add("hidden");
+	      const examples = examplesContainer.querySelector('.try_examples_content');
+	      iframe = document.createElement('iframe');
+	      iframe.src = iframeSrc;
+	      iframe.style.width = '100%';
+	      iframe.style.height = `${examples.offsetHeight}px`;
+	      iframe.classList.add('jupyterlite_sphinx_raw_iframe');
+	      examplesContainer.classList.add("hidden");
+	      iframeContainer.appendChild(iframe);
+    } else {
+	      examplesContainer.classList.add("hidden");
     }
     iframeParentContainer.classList.remove("hidden");
 }
--- a/jupyterlite_sphinx/jupyterlite_sphinx.js
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.js
@@ -48,7 +48,7 @@ window.tryExamplesShowIframe = (examplesContainerId, iframeContainerId, iframeSr
 
     iframe.src = iframeSrc;
     iframe.width = iframe.height = '100%';
-    iframe.classList.add('try_examples_raw_iframe');
+    iframe.classList.add('jupyterlite_sphinx_raw_iframe');
     examples_container.classList.add("hidden")
     iframe_container.appendChild(iframe);
     iframe_container.classList.remove("hidden")
--- a/jupyterlite_sphinx/jupyterlite_sphinx.js
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.js
@@ -41,15 +41,34 @@ window.jupyterliteConcatSearchParams = (iframeSrc, params) => {
 }
 
 
-window.tryExamplesShowIframe = (examplesContainerId, iframeContainerId, iframeSrc) => {
-    const examples_container = document.getElementById(examplesContainerId)
+window.tryExamplesShowIframe = (
+    examplesContainerId, iframeContainerId, iframeParentContainerId, iframeSrc
+) => {
+    const examples_container = document.getElementById(examplesContainerId);
+    const iframe_parent_container = document.getElementById(iframeParentContainerId);
     const iframe_container = document.getElementById(iframeContainerId);
-    const iframe = document.createElement('iframe');
 
-    iframe.src = iframeSrc;
-    iframe.width = iframe.height = '100%';
-    iframe.classList.add('jupyterlite_sphinx_raw_iframe');
-    examples_container.classList.add("hidden")
-    iframe_container.appendChild(iframe);
-    iframe_container.classList.remove("hidden")
+    let iframe = iframe_container.querySelector('iframe.jupyterlite_sphinx_raw_iframe');
+
+    if (!iframe) {
+	iframe = document.createElement('iframe');
+	iframe.src = iframeSrc;
+	iframe.width = iframe.height = '100%';
+	iframe.classList.add('jupyterlite_sphinx_raw_iframe');
+	examples_container.classList.add("hidden")
+	iframe_container.appendChild(iframe);
+    }
+    else {
+	examples_container.classList.add("hidden");
+    }
+    iframe_parent_container.classList.remove("hidden");
+}
+
+
+window.tryExamplesHideIframe = (examplesContainerId, iframeParentContainerId) => {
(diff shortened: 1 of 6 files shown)


jtpio:
Thanks @steppi!



> At least for Firefox, I've found that my browser prompts me to select a kernel every time I open one of the generated notebooks, even though Pyodide is the only kernel installed. You can try that out [here](https://steppi.github.io/scipy/reference/index.html). I think this will be too annoying for users, so hope there is a way the kernel can be specified in advance.

Agree the popup can be quite annoying. Maybe it's an issue in JupyterLite (or JupyterLab), which is not able to pick the default (and only) kernel when available.

Looking at one of the example notebooks: https://raw.githubusercontent.com/steppi/scipy/gh-pages/lite/files/004e9741_32fd_4b41_9716_052cde466244.ipynb, it looks like the `metadata` field of the notebook is empty. But adding `python` or the Pyodide kernel to the notebook metadata may also help JupyterLite pick the correct kernel on startup.

steppi:
Thanks so much @jtpio! Adding `python` to the notebook metadata worked.

steppi:
https://steppi.github.io/scipy/ as been updated with the fix and now doesn't have the annoying pop up to select the kernel. 

One piece of feedback I've received recently is that the button to try the examples should be at the top of the Examples section to make it more noticeable.

I noticed that when pressing the try examples button, you just get a blank white iframe for a second or two until the notebook loads. I think I need to add some kind of indication that the notebook is loading, like you do for the existing directives. I'll also work on adding documentation and more configurability.

martinRenou:
That looks great! Thanks!

> I'll also work on adding documentation and more configurability

Would you like to get a review already or should we wait for another iteration from your side?

steppi:
> That looks great! Thanks!
> 
> > I'll also work on adding documentation and more configurability
> 
> Would you like to get a review already or should we wait for another iteration from your side?

Thanks! It would be great if you could give it a first pass, even if just a quick one, just to help save me from putting in extra work finishing things up along a path that shouldn’t have been taken, if something isn’t quite right.

martinRenou:
> You can find this deployed for the SciPy documentation here, https://steppi.github.io/scipy/

Following the link your provided, I cannot test this feature. Do you have a link to a particular page in the documentation where I can try it?

> This works pretty well in most cases, but looks bad for particularly short Examples sections

Maybe an harcoded (or configurable) minimum size for the IFrame could improve this?

steppi:
> Following the link your provided, I cannot test this feature. Do you have a link to a particular page in the documentation where I can try it?

Sorry, here's an example page for the function [dblquad](https://steppi.github.io/scipy/reference/generated/scipy.integrate.dblquad.html#scipy.integrate.dblquad). If you want to see more, these interactive docstrings are in the API reference section of the docs.


> Maybe an harcoded (or configurable) minimum size for the IFrame could improve this?

Yeah, I think that's a good idea. I think we can make it configurable with a sane default.

martinRenou:
> Sorry, here's an example page for the function [dblquad](https://steppi.github.io/scipy/reference/generated/scipy.integrate.dblquad.html#scipy.integrate.dblquad). If you want to see more, these interactive docstrings are in the API reference section of the docs.

Neat 👍🏽

jupyterlite_sphinx/jupyterlite_sphinx.js
```diff
@@ -39,3 +39,38 @@ window.jupyterliteConcatSearchParams = (iframeSrc, params) => {
...
+	iframe.style.width = '100%';
+	iframe.style.height = `${examples.offsetHeight}px`;
+	iframe.classList.add('jupyterlite_sphinx_raw_iframe');
+	examplesContainer.classList.add("hidden");
+	iframeContainer.appendChild(iframe);
+    }
+    else {
+	examplesContainer.classList.add("hidden");
+    }
+    iframeParentContainer.classList.remove("hidden");
```
martinRenou, line 66: Small indent nitpicking

```suggestion
    if (!iframe) {
	      const examples = examplesContainer.querySelector('.try_examples_content');
	      iframe = document.createElement('iframe');
	      iframe.src = iframeSrc;
	      iframe.style.width = '100%';
	      iframe.style.height = `${examples.offsetHeight}px`;
	      iframe.classList.add('jupyterlite_sphinx_raw_iframe');
	      examplesContainer.classList.add("hidden");
	      iframeContainer.appendChild(iframe);
    } else {
	      examplesContainer.classList.add("hidden");
    }
    iframeParentContainer.classList.remove("hidden");
```

martinRenou (commented):
This all look sensible to me. 

I'm just afraid that manually parsing the rst like this would not scale very well and may be difficult to maintain.
I wonder if we can use `docutils`'s parsers or write our own parser inheriting from the base ones.
Anyway, I don't think this should be a blocker to this PR for now. We may want to turn this comment into an issue once the PR is merged.

steppi:
> This all look sensible to me.
> 
> I'm just afraid that manually parsing the rst like this would not scale very well and may be difficult to maintain. I wonder if we can use `docutils`'s parsers or write our own parser inheriting from the base ones. Anyway, I don't think this should be a blocker to this PR for now. We may want to turn this comment into an issue once the PR is merged.

Yeah, that makes sense. Scalability seems surprisingly Ok, SciPy is probably among the top packages for amount of docstring content, and the impact on the already enormous doc build times wasn't terrible. 100% agree on this manual parsing being more difficult to maintain and just overall more brittle. I didn't know how to make it work with an existing parser so just kind of went for it with the manual approach. Thanks for not considering this a blocker, and good idea to make an issue.

jupyterlite_sphinx/jupyterlite_sphinx.js
steppi: Thanks. New to Javascript so not really aware of the best ways to format things.

martinRenou:
> SciPy is probably among the top packages for amount of docstring content, and the impact on the already enormous doc build times wasn't terrible

Yep that's very good!

One thing I'm wondering is how it would work for documentation that's written in Markdown using something like myst-parser. Can we expect this code to work in that case?

Happy to merge this as-is now and we can iterate on it :)

martinRenou (approved):
Thank you! 🚀

Issue #110, opened 2023-09-14
Update to JupyterLite 0.2.0


martinRenou:
Which is now based on JupyterLab 4 and Notebook 7 packages: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.2.0a0

PR #113, opened 2023-11-08
Update to jupyterlite v0.2

martinRenou:
Fix https://github.com/jupyterlite/jupyterlite-sphinx/issues/110


docs/directives/notebooklite.md
```diff
@@ -0,0 +1,33 @@
+# NotebookLite directive
+
+`jupyterlite-sphinx` provides a `notebooklite` directive that allows you to embed the classic Notebook UI in your docs.
+
+```rst
+.. notebooklite::
```
jtpio, line 6: Maybe just `notebook` could be enough? We don't make any use of `notebooklite` elsewhere.

jtpio: Or maybe using an `:app:` parameter could help make this scale better in the future in case more apps are added?

Something like the following, using `notebooks` to match the name of the app in JupyterLite:

```rst
.. jupyterlite::
   :width: 100%
   :height: 600px
   :app: notebooks
   :prompt: Try classic Notebook!
```

jtpio: Yes, just opened https://github.com/jupyterlite/jupyterlite-sphinx/issues/114

PR #118, opened 2024-01-10
Add option to disable TryExamples without rebuilding docs

steppi:
This PR makes it so if a file entitled `.disable_try_examples` is added to the root level of the *deployed* documentation's *build* directory, "try examples" buttons will be hidden at page load time, effectively disabling interactive examples without requiring a documentation rebuild.

A Javascript function checks at page load for the presence of the file, and if so, hides the buttons.

In https://github.com/scipy/scipy/issues/19729#issuecomment-1875437428, @rgommers noted that it would be good to be able to disable the interactive examples without requiring a rebuild of the documentation. This is particularly helpful for SciPy with its long documentation build and deployment times.

image

### DIFF
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -167,16 +167,16 @@ If you are using the `TryExamples` directive in your documentation, you'll need 
 that the version of the package installed in the Jupyterlite kernel you are using
 matches that of the version you are documenting.
 
-## Disable without rebuilding
+## Configuration without rebuilding
 The `TryExamples` directive supports disabling interactive examples without rebuilding
 the documentation. This can be helpful for projects requiring substantial documentation
-build time. Users may add a json file entitled `.try_examples_ignore.json` to the root
+build time. Users may add a json config file entitled `.try_examples.json` to the root
 directory of the build directory for the deployed documentation. The format is a list of
-[JavaScript Regex patterns](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions) attached to the key `"patterns"` like below.
+[JavaScript Regex patterns](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions) attached to the key `"ignore_patterns"` like below.
 
 ```json
 {
-    "patterns": ["^latest/.*", "^stable/reference/generated/example"]
+    "ignore_patterns": ["^latest/.*", "^stable/reference/generated/example"]
 }
 ```
 
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -166,3 +166,9 @@ allowing for specification of examples sections which should not be made interac
 If you are using the `TryExamples` directive in your documentation, you'll need to ensure
 that the version of the package installed in the Jupyterlite kernel you are using
 matches that of the version you are documenting.
+
+## Disable without rebuilding
+Adding a file entitled `.disable_try_examples` to the root of your documentation's build
+directory will cause try examples buttons to be hidden at page load time, effectively
+disabling interactive examples without requiring a documentation rebuild. This can be
+helpful for projects requiring substantial documentation build time.
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -195,3 +195,12 @@ Note that these patterns should match the Sphinx docnames of the documentation f
 A docname is the relative path from the documentation root to the file of interest,
 but without the file extension. For instance, the docname corresponding to `index.rst` at
 the root of the Sphinx source directory would be `index`.
+
+A default configuration file can be specified in `conf.py` with the option
+`try_examples_default_runtime_config`.
+
+```python
+try_examples_default_runtime_config = {
+    "ignore_patterns": ["^latest/.*", "^stable/reference/generated/example"]
+}
+```
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -168,7 +168,30 @@ that the version of the package installed in the Jupyterlite kernel you are usin
 matches that of the version you are documenting.
 
 ## Disable without rebuilding
-Adding a file entitled `.disable_try_examples` to the root of your documentation's build
-directory will cause try examples buttons to be hidden at page load time, effectively
-disabling interactive examples without requiring a documentation rebuild. This can be
-helpful for projects requiring substantial documentation build time.
+The `TryExamples` directive supports disabling interactive examples without rebuilding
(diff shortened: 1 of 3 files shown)


jupyterlite_sphinx/jupyterlite_sphinx.js
```diff
@@ -76,3 +76,18 @@ window.tryExamplesHideIframe = (examplesContainerId, iframeParentContainerId) =>
...
+    const disableFileUrl = `${relativePathToRoot}/.disable_try_examples?cb=${timestamp}`
+    fetch(disableFileUrl).then(response => {
+	if (response.ok) {
+            var buttons = document.getElementsByClassName('try_examples_button');
+            for (var i = 0; i < buttons.length; i++) {
+		buttons[i].classList.add('hidden');
+            }
+	}
+    });
+}
```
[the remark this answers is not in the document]
martinRenou: https://www.youtube.com/watch?v=SsoOG6ZeyUI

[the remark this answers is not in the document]
steppi: Ha. I use Emacs and for languages I use regularly, I have it set up to handle the indentation automatically. I guess the out of the box Javascript mode isn’t doing that properly. I’ll look into installing proper tooling.

steppi:
> To make thinks a bit more automatic, I would take a peak at the URL and look at wether it contains `/latest`|`/stable` or similar, or at least something the file you look at could be a json with a list of url pattern where this should be enabled/disabled.
> 
> At least the `latest/stable` would allow to have automatic enabling/disabling automatically when new version is pushed on RTD

Having it be a json with a list of url patterns sounds like a good solution.

steppi:
Thanks @Carreau! I've updated it to check for a json file `.try_examples_ignore.json` containing a list of regex patterns for files where interactive documentation should be disabled. I've been testing this locally, but can work on getting it set up for my experimental deployment at https://steppi.github.io/scipy/. A hiccup there is that I accidentally left circle ci enabled in my SciPy fork after building and deploying the docs the last time and ran out of credits for the month, so it's not as simple as just pushing a new commit where `jupyterlite-sphinx` is pinned to this PR branch.

Another thought I had is that it would be useful to be able to specify a default `.try_examples_ignore.json` in Sphinx's `conf.py`.

steppi:
I've deployed the docs at https://steppi.github.io/scipy with jupyterlite-sphinx pinned to this PR branch and added an ignore file https://github.com/steppi/scipy/blob/gh-pages/.try_examples_ignore.json. 

An example of a linalg example which is disabled: https://steppi.github.io/scipy/reference/generated/scipy.linalg.solve.html#scipy.linalg.solve

An example of a non-linalg example which is still live: https://steppi.github.io/scipy/reference/generated/scipy.cluster.hierarchy.linkage.html#scipy.cluster.hierarchy.linkage.

steppi:
OK, I've changed `.try_examples_ignore.json` to just a generic config file `.try_examples.json` and documented it as such, since we might want to add more runtime changeable options in the future. I've also made it possible to specify a default version of this file in the sphinx `conf.py`.

Carreau:
From a discussion, maybe we should have a "force-enable" function in the browser javascript console to force-enable.

docs/directives/try_examples.md
```diff
@@ -166,3 +166,41 @@ allowing for specification of examples sections which should not be made interac
 If you are using the `TryExamples` directive in your documentation, you'll need to ensure
 that the version of the package installed in the Jupyterlite kernel you are using
 matches that of the version you are documenting.
+
+## Configuration without rebuilding
```
Carreau, line 170:
```suggestion
## Configuration without rebuilding

```

Just a new line for readability.

docs/directives/try_examples.md
```diff
@@ -166,3 +166,41 @@ allowing for specification of examples sections which should not be made interac
...
+[JavaScript Regex patterns](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions) attached to the key `"ignore_patterns"` like below.
+
+```json
+{
+    "ignore_patterns": ["^latest/.*", "^stable/reference/generated/example"]
+}
+```
+
+`TryExamples` buttons will be hidden in files matching at least one of these patterns,
+effectively disabling the interactive documentation. In the provided example:
```
Carreau, line 184: Do we want those to match files, or final URLs ?

jupyterlite_sphinx/jupyterlite_sphinx.js
```diff
@@ -76,3 +76,43 @@ window.tryExamplesHideIframe = (examplesContainerId, iframeParentContainerId) =>
...
+                    for (var i = 0; i < buttons.length; i++) {
+                        buttons[i].classList.add('hidden');
+                    }
+                    break;
+                }
+            }
+        })
+        .catch(error => {
+            console.error(error);
+        });
```
Carreau, line 117: I talk to some local colleagues, and they told that most recent browser support top level await, so I think we canlater refactor into clearer code and avoid promises.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -530,7 +530,18 @@ def run(self):
...
 
-        return [content_container_node, notebook_container, style_tag]
+        # Search cnofig file allowing for config changes without rebuilding docs.
+        config_path = os.path.join(relative_path_to_root, ".try_examples.json")
+        script_html = (
+            "<script>"
+            'document.addEventListener("DOMContentLoaded", function() {'
+            f'window.loadTryExamplesConfig("{config_path}","{docname}");'
+            "});"
+            "</script>"
```
Carreau, line 540: I think we want the regex to match urls, that way if say on readthedocs we can have only `/stable/` and `/latest/` to show JupyterLight buttons, but not older version.

steppi: Ah, true. Mix up on my part. I didn't consider that we'd have two sets of deployed docs with different doc roots one at `base_url +  '/stable'` and the other at `base_url + '/latest'`. I was thinking of the case where `/` is the doc root, and it has subfolders `/stable` and `/latest`, which isn't how things would typically work.

steppi:
@Carreau, I've updated this to match url pathnames instead of sphinx docnames; use async/await instead of promises, and have added a function to toggle the visibility of the buttons for use from the console.

Carreau:
Thanks, I think we should just merge and iterate, to see how it goes.

martinRenou:
Thank you both!

steppi:
Thanks @Carreau! I'll work on implementing your other suggestions today.

Issue #127, opened 2024-01-24
:button_css: option seem to not only affect the current button.


Carreau:
my understanding was that :

```
.. try_examples::
    :button_css:
        background-color: red;
```

Would only make the *current* button red, though it seem to affect at least all the buttons on the page.

Maybe related to #122, but I suggest we have a single `:example-class: <str>`, that can be styled by the custom css. Even the button-top/button-bottom can be styled using custom flex-column/flex-reverse-column.

Carreau:
Yeah, I think an `:example-class:` or `:example-id:` as technically if something is unique in css you should use ID, but I think class + custom css should be able to do anything if we put the class on the outer try-example div.

Carreau:
I think this is closed by #129.

PR #129, opened 2024-01-26
Improve TryExamples customization

### LINKED ISSUE #122: Rework the order of "try-example" documentation
I think that the way the documentation for "try_examples" is written, it implicitely state that you _must_ include `button_css` and `button_css_hover` when creating a directive. 

I think it would be better to :
- have  minimal example at the top of the page, and
- list all options later in the page. 

At least I think that the option for global CSS should be listed _before_ individual button overwrite. 

I also think that the options for custom css are an anti pattern, We should rely on existing mechanism, that is to say [custom css and/or js file](https://docs.readthedocs.io/en/stable/guides/adding-custom-css.html#how-to-add-custom-css-or-javascript-to-sphinx-documentation), and simply document that classes that the button have depending the directive options. 

Having the style be in an actual css file allow all the linting/formatting/ci tools to work, as well as allow feature  like viewport dependant style, that can't be done with inlining css in the conf file.

steppi:
Closes #122, #127

This PR updates `TryExamplesDirective` to use the standard method of allowing custom css in Sphinx to customize the buttons as suggested by Carreau [here](https://github.com/jupyterlite/jupyterlite-sphinx/issues/122#issue-2081634955). It also adds customization of `min_height` to the config file `try_examples.json` so that the minimum height can bet changed at runtime. An option has been given to the directive to set the specific height of the iframe container for the embedded notebook and the `min_height` option has been removed. The specific height overrides the minimum height set in `try_examples.json`. I've also attempted to improve the documentation, and have added an `example_class` option to the directive which can be targeted in custom css as suggested by Carreau.

### DIFF
--- a/docs/_static/try_examples.css
+++ b/docs/_static/try_examples.css
@@ -1,0 +1,21 @@
+.try_examples_button {
+    background-color: #f7dc1e;
+    border: none;
+    padding: 5px 10px;
+    border-radius: 15px;
+    font-family: vibur;
+    font-size: larger;
+    box-shadow: 0 2px 5px rgba(108,108,108,0.2);
+}
+
+.try_examples_button:hover {
+    background-color: #fff221;
+    transform: scale(1.02);
+    box-shadow: 0 2px 5px rgba(0, 0, 0, 0.2);
+    cursor: pointer;
+}
+
+.try_examples_button_container {
+    display: flex;
+    justify-content: flex-end;
+}
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,4 +28,7 @@ html_theme_options = {
     ]
 }
 
+html_static_path = ["_static"]
+html_css_files = ["try_examples.css"]
+
 suppress_warnings = ["myst.xref_missing"]
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -1,34 +1,20 @@
 # TryExamples directive
 
 `jupyterlite-sphinx` provides the experimental `try_examples` directive which allows
-docstring examples sections written in [doctestformat](https://docs.python.org/3/library/doctest.html) to be swapped with an embedded classic Notebook at the push of a button.
+docstring examples sections written in [doctest format](https://docs.python.org/3/library/doctest.html) to be swapped with an embedded classic Notebook at the push of a button.
+
+Below is an example of the directive in use. The button has been styled with custom
+css as explained in the configuration section below. Without custom css, the button will
+be plain and unadorned.
 
 
 ```rst
 Examples
 --------
 .. try_examples::
-    :button_css:
-        background-color: #f7dc1e;
-	border: none;
-    	padding: 5px 10px;
-   	border-radius: 15px;
-	font-family: vibur;
-	font-size: x-large;
-	box-shadow: 0 2px 5px rgba(108,108,108,0.2);
-    :button_hover_css:
-        background-color: #fff221;
- 	transform: scale(1.02);
-	box-shadow: 0 2px 5px rgba(0, 0, 0, 0.2);
-	cursor: pointer;
-    :button_horizontal_position: right
-    :button_vertical_position: top
-    :button_text: Try it in a classic notebook!
-    :min_height: 200px
-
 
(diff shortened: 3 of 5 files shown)


docs/_static/try_examples.css
```diff
@@ -0,0 +1,21 @@
...
+    background-color: #fff221;
+    transform: scale(1.02);
+    box-shadow: 0 2px 5px rgba(0, 0, 0, 0.2);
+    cursor: pointer;
+}
+
+.try_examples_button_container {
+    display: flex;
+    justify-content: flex-end;
+}
```
Carreau, line 21: :+1: Thanks for your patience on the review, I'm just back from PTO.

docs/conf.py
```diff
@@ -28,4 +28,7 @@
     ]
 }
 
+html_static_path = ["_static"]
+html_css_files = ["try_examples.css"]
+
```
Carreau, line 33: 👍

docs/directives/try_examples.md
```diff
@@ -104,75 +75,144 @@ Examples
...
+    transform: scale(1.02);
+    box-shadow: 0 2px 5px rgba(0, 0, 0, 0.2);
+    cursor: pointer;
+}
+
+.try_examples_button_container {
+    display: flex;
+    justify-content: flex-end;
+}
+```
```
Carreau, line 111: 👍

docs/directives/try_examples.md
```diff
@@ -184,23 +224,38 @@ directory of the build directory for the deployed documentation. The format is a
 `TryExamples` buttons will be hidden in url pathnames matching at least one of these
 patterns, effectively disabling the interactive documentation. In the provided example:
 
-* The pattern `".*latest/.*" disables interactive examples for urls for the documentation
+* The pattern `"^/latest/.*"` disables interactive examples for urls for the documentation
```
Carreau, line 227: good catch.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -368,13 +364,10 @@ class TryExamplesDirective(SphinxDirective):
...
     option_spec = {
+        "height": directives.unchanged,
         "theme": directives.unchanged,
         "button_text": directives.unchanged,
-        "button_css": directives.unchanged,
-        "button_hover_css": directives.unchanged,
-        "button_horizontal_position": directives.unchanged,
-        "button_vertical_position": directives.unchanged,
-        "min_height": directives.unchanged,
+        "example_class": directives.unchanged,
```
Carreau, line 370: 👍

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -387,25 +380,8 @@ def run(self):
...
-                f" 'left', 'center', or 'right', received {button_horizontal_position}."
-            )
-
-        if button_vertical_position not in ["bottom", "top"]:
-            raise RuntimeError(
-                "try_examples directive expects button_vertical_position to be one of"
-                f" 'top' or 'bottom', received {button_vertical_position}."
-            )
+        height = self.options.pop("height", None)
+        example_class = self.options.pop("example_class", "")
```
Carreau, line 384: Not a fix for this PR, but do we want to check here that `example_class` is a valid class name ? (no space, no dots, no /\ or angle brackets for example) as we do inject it with fstrings it might break the html/css otherwise

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -689,27 +607,11 @@ def setup(app):
...
     app.add_config_value(
         "try_examples_global_button_text",
         default=None,
         rebuild="html",
     )
-    app.add_config_value(
-        "try_examples_default_runtime_config",
-        default=None,
-        rebuild=None,
-    )
```
Carreau, line 712: I love when a bunch of lines get deleted.

Carreau (approved):

Carreau:
Ok, let's get that in. I'll open an issue to think about validating the height and example_class

steppi:
Thanks @Carreau!

PR #131, opened 2024-01-31
Misc parsing warnings.

Carreau:
Here are a couple of modification:

The example_class was not applied to all the container.

Having both the default and iframe container with the same class was a bit annoying for some css, and lead to `.hidden` not being applied in some of my experiments, so swap the name of the iframe container.

Add an example in the documentation on how to change the color and positioning of the button.

I happen to have prettier install for css files, so it also reformatted the css to use css standard of 2 spaces.

Move the main colors into css variables, so that if need and if we start to use those colors in more places they are more easily overwritten.

Also contains mist parsing warnings.

 - Unknown escape sequence \;, need double escape.
 - pygments say yml is unknown but yaml works

### DIFF
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -125,6 +125,7 @@ Here's an example with some options set
 .. try_examples::
     :button_text: Try it in your browser!
     :height: 400px
+    :example_class: blue-bottom
 
     The button text has changed and the height now exceeds the size of the content.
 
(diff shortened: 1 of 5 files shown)


docs/directives/try_examples.md
```diff
@@ -133,6 +134,11 @@ Here's an example with some options set
     >>> x + y
     4
 
+    We've also added the ``blue-bottom`` class, the button should appear as blue,
+    below the examples, and on the left side of the screen.
+
+    See ``try_examples.css`` in the Repository to see how we achieved this via
+    custom css.
```
steppi, line 141:
```suggestion
    See `try_examples.css <https://github.com/jupyterlite/jupyterlite-sphinx/blob/main/docs/_static/try_examples.css>`_ to see how we achieved this via custom css.
```

I think it would be helpful to make this a link to the file.

steppi (approved):
Thanks @Carreau. Looks good modulo adding a link to `try_examples.css` in the docs.

steppi:
I've gone ahead and merged this and will add the link to `try_examples.css` in a follow-up.

PR #133, opened 2024-01-31
Add a loading spinner for TryExamples directive.

### LINKED ISSUE #120: Add a spinner "Try example directive"
The first time you click on the "Try example" directive; especially on a slow connection, the iframe can stay blank for quite some time. 

It would be good to inject a spinner like the `..jupyterlite` directive that says "loading...", to make sure the user is not refreshing or wondering what is happening.

steppi:
Closes #120

This PR adds a loading spinner for the TryExamples directive. I've offered no options to customize the spinner because I don't have sufficient time to think about how to do that cleanly and document it properly. My goal is just to get this working before submitting a PR adding interactive docs to SciPy. We can revisit later.

I've also fixed a bug where I was referring to a configuration variable which no longer exists (it only occurs when `global_enable_try_examples = True`), and I've added a link to `try_examples.css` in the try examples documentation to make it easier for readers to find that file.

### DIFF
--- a/jupyterlite_sphinx/jupyterlite_sphinx.css
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.css
@@ -68,11 +68,6 @@
     }
 }
 
-.try_examples_iframe_container {
-    position: relative;
-    cursor: pointer;
-}
-
 
 .try_examples_outer_container {
     position: relative;
@@ -81,4 +76,24 @@
 
 .hidden {
     display: none;
+}
+
+
+.try_examples_spinner {
+    /* From https://css-loaders.com/spinner/ */
+    position: absolute;
+    z-index: 0;
+    top: 20%;
+    left: 50%;
+    width: 50px;
+    aspect-ratio: 1;
+    border-radius: 50%;
+    background:
+        radial-gradient(farthest-side,#ffa516 94%,#0000) top/8px 8px no-repeat,
+        conic-gradient(#0000 30%,#ffa516);
+    -webkit-mask: radial-gradient(farthest-side,#0000 calc(100% - 8px),#000 0);
+    animation: l13 1s infinite linear;
+}
+@keyframes l13{
+    100%{transform: rotate(1turn)}
 }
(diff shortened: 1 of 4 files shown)


steppi:
I've deployed SciPy's docs built against this feature branch. You can check out the spinner in action here https://steppi.github.io/scipy/reference/generated/scipy.stats.kurtosistest.html#scipy.stats.kurtosistest

jupyterlite_sphinx/jupyterlite_sphinx.css
```diff
@@ -82,3 +77,22 @@
 .hidden {
     display: none;
 }
+
+
+.try_examples_spinner {
```
martinRenou, line 82: It would be nice to generalize this across jupyterlite-sphinx. For example, having this spinner show up instead of the "Loading..." message when clicking on "Try JupyterLite" here: https://jupyterlite-sphinx.readthedocs.io/en/latest/directives/jupyterlite.html

martinRenou (changes requested):
I tried the link you pointed to, looks great!!

It seems the spinner is not centered on the iframe, which is odd given the CSS you apply (absolute positioning + width 50% and height 50%), not sure what's wrong.

jupyterlite_sphinx/jupyterlite_sphinx.js
```diff
@@ -62,9 +67,16 @@ window.tryExamplesShowIframe = (
               } else {
                   height = Math.max(tryExamplesGlobalMinHeight, examples.offsetHeight);
               }
+
+              // Get spinner position
+              const iframeTop = iframe.getBoundingClientRect().top;
+              const spinnerTop = iframeTop + Math.min((height * 0.2), 200);
+              spinner.style.top = `${spinnerTop}px`;
```
steppi, line 74: > It seems the spinner is not centered on the iframe, which is odd given the CSS you apply (absolute positioning + width 50% and height 50%), not sure what's wrong.

The spinner position gets set in the javascript here. I put it closer to the top because a lot of SciPy's examples are very long, and a centered spinner usually isn't visible when the button is still in view in those cases. I guess it would make more sense to set it to be 50% unless this ends up out of the viewport, and otherwise put just close enough to the top to be in view.

steppi:
SciPy docs incorporating the updates are up now: https://steppi.github.io/scipy/reference/generated/scipy.stats.kurtosistest.html#scipy.stats.kurtosistest

martinRenou (approved):
Thanks!

PR #137, opened 2024-02-13
Suggestion: Add pre-commit to format js and css files.

Carreau:
I don't believe most of us know the current conventions and have the tooling installed. So it might be a good thing to have a consistent style.

If we do this we can also add pre-commit-ci so that PR are automatically updated. I'm generally not a fan of pre-commit hooks in general and pre commit-ci as it either can take forcer before you can already commit, or also when CI push on the PR, prevent pushing before pulling which can be annoying for contributor. So either choice is good for me.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,7 @@
+repos:
+- repo: https://github.com/pre-commit/mirrors-prettier
+  rev: 'f12edd9'  # Use the sha / tag you want to point at
+  hooks:
+  - id: prettier
+    types_or: [css, javascript]
+


Carreau:
Any thought on this ? I keep forgetting it's not using prettier, and my editor keep reformatting the whole file, so I would to either get that in, or know it won't so I can look into configuring my system to not format JS only on this project.

jtpio:
Sounds good, thanks!

As a follow-up, maybe we can also enable pre-commit CI for the repo?

Carreau:
Yes, +1 on pre-commit-ci, it slightly annoying sometime when it pushed and you need to pull/push, but I think it's better in most cases.

Carreau:
Ok, let's get that in. I'll add pre-commit-ci later.

PR #138, opened 2024-02-23
Update to `jupyterlite-xeus`

jtpio:
`jupyterlite-xeus` is the new way for loading xeus kernels: https://jupyterlite-xeus.readthedocs.io/en/latest/

`jupyterlite-xeus-python` is now archived: https://github.com/jupyterlite/xeus-python-kernel

We should reflect that in the documentation.


martinRenou (approved):
Thanks!

PR #139, opened 2024-02-28
Fix incorrect math processing

### LINKED ISSUE #136: problem with math directive in try example
It seem math direcive as seen in https://jupyterlite-sphinx.readthedocs.io/en/latest/directives/try_examples.html seem to me missparsed.

it ends up in the notebook as 

````
.. math

```
<the equations>
```
````

Note the missing `::`.

Screenshot 2024-02-13 at 10 41 34

steppi:
Closes #136

This PR is an attempt to fix the incorrect math processing noted by @Carreau in #136.. It seems like the problem was introduced in https://github.com/jupyterlite/jupyterlite-sphinx/pull/134 which added handling for literal blocks. Processing `.. math::` blocks before literal blocks appears to fix the problem.

I'll build SciPy's documentation against this PR and check if there are any unintended consequences.


Carreau (approved):
Seem to work for me, let's merge and iterate.

Issue #140, opened 2024-02-29
Hide button, or warn about high data download on mobile.


Carreau:
See [19729](https://github.com/scipy/scipy/issues/19729#issuecomment-1969621588) for context, we might want to have option/theme to not show button on mobile, as also Jupyter Interface is not mobile friendly.

PR #141, opened 2024-02-29
Hide buttons on smaller screens (mobile).

Carreau:
See #140, we might want to also add a warning dialog from Javascript as well in a subsequent PR.

### DIFF
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -7,6 +7,11 @@ Below is an example of the directive in use. The button has been styled with cus
 css as explained in the configuration section below. Without custom css, the button will
 be plain and unadorned.
 
+Note that as starting JupyterLite can download a significant amount of data, and
+that the Jupyter interface is not optimized for mobile, the buttons will be
+hidden on mobile by default (screen width 430px or smaller). This can be
+changed by overwriting with custom CSS.
+
 
 ```rst
 Examples
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -9,7 +9,7 @@ be plain and unadorned.
 
 Note that as starting JupyterLite can download a significant amount of data, and
 that the Jupyter interface is not optimized for mobile, the buttons will be
-hidden on mobile by default (screen width 430px or smaller). This can be
+hidden on mobile by default (screen width 480px or smaller). This can be
 changed by overwriting with custom CSS.
 
 
--- a/jupyterlite_sphinx/jupyterlite_sphinx.css
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.css
@@ -70,7 +70,7 @@
 /* we do not want the button to show on smaller screens (phones), as clicking
  * can download a lot of data, 430px, is the width of iPhone 14 pro max */
 
-@media (max-width: 430px) {
+@media (max-width: 480px), (max-height: 480px) {
   div.try_examples_button_container {
     display: none;
   }
--- a/jupyterlite_sphinx/jupyterlite_sphinx.css
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.css
@@ -68,7 +68,7 @@
 
 
 /* we do not want the button to show on smaller screens (phones), as clicking
- * can download a lot of data, 430px, is the width of iPhone 14 pro max */
+ * can download a lot of data. 480px is a commonly used breakpoint to identify if a device is a smartphone.  */
 
 @media (max-width: 480px), (max-height: 480px) {
   div.try_examples_button_container {
--- a/jupyterlite_sphinx/jupyterlite_sphinx.css
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.css
@@ -64,3 +64,14 @@
 @keyframes l13{
     100%{transform: rotate(1turn)}
 }
+
+
+
+/* we do not want the button to show on smaller screens (phones), as clicking
+ * can download a lot of data, 430px, is the width of iPhone 14 pro max */
+
+@media (max-width: 430px) {
+  div.try_examples_button_container {
+    display: none;
+  }
+}


mattip:
Thanks!

martinRenou (commented):
Should we put this behind a jupyterlite-sphinx flag?

Carreau:
> Should we put this behind a jupyterlite-sphinx flag?

I don't think we should this can always be overridden with custom css, the same way we can change the buttons colors or position on the page.

martinRenou:
Ok, sounds good!

steppi:
@martinRenou After this gets in, can we do a new release?

steppi:
> Sure thing! You should have the rights to make releases. Feel free to make it if you want!
> 
> To do this you need to go at the `Actions` tab of the repo, then manually trigger the `Step 1: Prep Release`. This should publish a draft release in https://github.com/jupyterlite/jupyterlite-sphinx/releases once done. Then you can manually run the `Step 2: Publish Release` to finalize it.

Thanks @martinRenou! That's very convenient. I'll make one after things get settled.

steppi:
It seems that the button is not hidden on my phone when I hold it horizontally

B39000D9-88F3-48DB-8AAF-253AD0E542BC

but is hidden when held vertically 

2AEA4AE6-FB58-4E37-A5AF-5A7498F3325F

steppi:
@Carreau, could we use javascript to conditionally hide the button based on the area of the screen? Or would it be better just to use plain css and hide based on max-width and max-height? Also, should we take tablets into account?

mattip:
> Also, should we take tablets into account?

Often tablets will be used in school/university settings where it would be helpful to be able to try the code. It is a shame there is no "low data connection" indicator, this seems like it should be a solved problem.

agriyakhetarpal:
Is it possible to, instead of tinkering with screen sizes, retrieve at the device settings and warn users with a pop-up / differently-coloured ~or greyed-out~ button that using the kernel will be slower on mobile devices?

steppi:
> > Also, should we take tablets into account?
> 
> Often tablets will be used in school/university settings where it would be helpful to be able to try the code. It is a shame there is no "low data connection" indicator, this seems like it should be a solved problem.

@Carreau mentioned [navigator.connection](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/connection#browser_compatibility) [here](https://github.com/scipy/scipy/issues/19729#issuecomment-1970793360) which should work in principle, but isn't available on ios safari.. It seems like they don't want to implement it because of privacy concerns. This issue, https://bugs.webkit.org/show_bug.cgi?id=185697, has been open for almost 6 years now.

steppi:
> Is it possible to, instead of tinkering with screen sizes, retrieve at the device settings and warn users with a pop-up / differently-coloured ~or greyed-out~ button that using the kernel will be slower on mobile devices?

Not sure. I think the issue of having a pop-up or changing the color of the button is orthogonal to detecting if we're dealing with a device using mobile data. @Carreau mentioned adding a warning dialogue with javascript, so I think that is the plan eventually. I don't know much about web dev, so I'm not sure what you mean by retrieving at the device settings. Based on the research I've done this morning, I think the only things we have to go on are the user agent, screen width, screen height, and on some platforms `navigator.connection`.  We could probably come up with a good heuristic that takes all of these things into account, but for now its probably best to start simple.

Carreau:
we can likely implement a dialog we a "don't ask me again", and tweak the logic of when to show it later.

docs/directives/try_examples.md
```diff
@@ -7,6 +7,11 @@ Below is an example of the directive in use. The button has been styled with cus
 css as explained in the configuration section below. Without custom css, the button will
 be plain and unadorned.
 
+Note that as starting JupyterLite can download a significant amount of data, and
+that the Jupyter interface is not optimized for mobile, the buttons will be
+hidden on mobile by default (screen width 430px or smaller). This can be
```
steppi, line 12:
```suggestion
hidden on mobile by default (screen width 480px or smaller). This can be
```

jupyterlite_sphinx/jupyterlite_sphinx.css
```diff
@@ -64,3 +64,14 @@
 @keyframes l13{
     100%{transform: rotate(1turn)}
 }
+
+
+
+/* we do not want the button to show on smaller screens (phones), as clicking
+ * can download a lot of data, 430px, is the width of iPhone 14 pro max */
```
steppi, line 71:
```suggestion
 * can download a lot of data. 480px is a commonly used breakpoint to identify if a device is a smartphone.  */
```

steppi (commented):
Looks good and works correctly on my phone now. Just needs some comments updated in response to the 430px to 480px change.

steppi (approved):
Looks good.

steppi:
@martinRenou, it seems like I might not have the right permissions to make a release. I tried running the `Step 1: Prep Release` action like you described above, and the action failed. The logs say

```
  File "/home/runner/work/_actions/jupyter-server/jupyter_releaser/v2/jupyter_releaser/actions/common.py", line 19, in setup
    return prepare_environment(fetch_draft_release=fetch_draft_release)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/_actions/jupyter-server/jupyter_releaser/v2/jupyter_releaser/util.py", line 618, in prepare_environment
    raise RuntimeError(msg) from None
RuntimeError: Could not get user permission level, assuming user was not admin!
```

steppi:
> @steppi can you try again?

It worked this time. Thanks!

steppi:
@martinRenou, I just selected the default options, which created release `0.11.1`, but I think this makes more sense as an `0.12.0`, since there were a number of enhancements, and having the `TryExamples` directive no longer work on small screens could be seen as backwards incompatible.

martinRenou:
Right, you can remove the release draft in https://github.com/jupyterlite/jupyterlite-sphinx/releases. Then I believe you can rerun with the version specifier "minor".

martinRenou:
Yay!! Thanks for the release!!

PR #143, opened 2024-02-29
Add option for initial warning cell in try examples directive

### LINKED ISSUE #142: Add option for warning in first cell of TryExamples generated notebook
### Problem

It would be helpful to be able to warn users about things like slow load times etc. In the notebooks Scikit-Learn uses with Jupyterlite, they put a warning in the first cell like this

image

### Proposed Solution

I'll submit a PR which adds a config option which specifies a warning text which can be added in the first cell.

steppi:
Closes #142 

This PR updates the code for auto generating the notebooks to have the option to add an initial "warning" cell like discussed in #142. This has been added as a configuration option both at the individual directive level, and at the global level with option `try_examples_global_warning_text`, similar to the existing `try_examples_global_button_text` etc.

I just looked the json for one of Scikit-Learn's notebooks and copied that formatting there. Basically, wrapping the text supplied in the config with a `<div class='alert alert-warning'>`.  Jupyter is apparently has some css that's aware of this formatting and styles things appropriately.

### DIFF
--- a/jupyterlite_sphinx/_try_examples.py
+++ b/jupyterlite_sphinx/_try_examples.py
@@ -3,7 +3,7 @@ from nbformat.v4 import new_code_cell, new_markdown_cell
 import re
 
 
-def examples_to_notebook(input_lines, warning_text=None):
+def examples_to_notebook(input_lines, *, warning_text=None):
     """Parse examples section of a docstring and convert to Jupyter notebook.
 
     Parameters
--- a/jupyterlite_sphinx/_try_examples.py
+++ b/jupyterlite_sphinx/_try_examples.py
@@ -48,13 +48,11 @@ def examples_to_notebook(input_lines, warning_text=None):
     """
     nb = nbf.v4.new_notebook()
 
-
     if warning_text is not None:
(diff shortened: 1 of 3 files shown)


jupyterlite_sphinx/_try_examples.py
```diff
@@ -3,13 +3,17 @@
 import re
 
 
-def examples_to_notebook(input_lines):
+def examples_to_notebook(input_lines, warning_text=None):
```
Carreau, line 6: Make it keyword only for simple future API changes ? 

```suggestion
def examples_to_notebook(input_lines, *, warning_text=None):
```

Carreau (approved):

steppi:
I've rebuilt the SciPy docs at https::/steppi.github.io/scipy against this branch. See https://steppi.github.io/scipy/reference/generated/scipy.integrate.quad.html#scipy.integrate.quad for a specific example.

Carreau:
Let's get that in. Thanks !

steppi:
Thanks @Carreau!

PR #144, opened 2024-03-05
Think of dialog to confirm whether to load JupyterLight

Carreau:
~~This is not perfect, in particular the `confirm(...)` is a blocking call, which is not supposed to be used in a OnClik handler as an onclick handler need to respond fast.~~

~~I think this need to be refactored into utilities functions, and setTimeout(0, callback), and/or Promises.~~

~~That way the onClick finishes immediately and does not violate the timeout limit.~~

We can't use SetTimeout either as it also suffer the same problem. For now I think we can live with the warning in the console on slow  or metered connections.

As noted in the comment, the navigator api is experimental, so it would be good to test properly. In Web Inspector  (Firefox and Chome at least), it is possible to artificially throttle the connection to test the logic.

### DIFF
--- a/jupyterlite_sphinx/jupyterlite_sphinx.js
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.js
@@ -52,6 +52,41 @@ window.jupyterliteConcatSearchParams = (iframeSrc, params) => {
   }
 };
 
+/*
+ * On slow or metered connection we might not want to download a few hundred of
+ * Mb without warnings, this will try to detect slow or metered connections and
+ * ask the user wheter they wish to continue.
+ *
+ * If we detect that the user is not on a metered connection or slow connection
+ * it will return true.
+ */
+const _askContinueToUserIfMobileOrSlow = () => {
+  // this is an experimental API
+  // https://developer.mozilla.org/en-US/docs/Web/API/Navigator/connection
+  const networkinfo = navigator.connection || {};
+  const connectiontype = networkinfo.effectiveType || "";
+  if (
+    networkinfo.saveData == true ||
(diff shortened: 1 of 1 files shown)


steppi:
I think the downloads only happen if someone tries to execute a cell with imports in it. I wonder if it would be sufficient to add a note about large downloads in the initial warning cell? My impression is Apple has no intention of ever implementing `navigator`, so only using`navigator` seems like it might be an unworkable solution in practice.

martinRenou:
> I think the downloads only happen if someone tries to execute a cell with imports in it.

Just adding a note that this is true for the pyodide kernel, not true for the xeus-python kernel which would download the Python environment at kernel startup. So the first cell approach does not fit the xeus-python case, because the kernel would have started as soon as you open the notebook.

steppi:
> > I think the downloads only happen if someone tries to execute a cell with imports in it.
> 
> Just adding a note that this is true for the pyodide kernel, not true for the xeus-python kernel which would download the Python environment at kernel startup. So the first cell approach does not fit the xeus-python case, because the kernel would have started as soon as you open the notebook.

Thanks, that’s good to know. I want to add the warning about large downloads when I deploy for SciPy’s docs, but it seems we still need a principled solution.

Carreau:
Sorry for the delay in responding. I think we can design the mechanism we wish to alert user and iterate over the API and predicate we use to who it. 

Note that the "download" buttons already should be hidden on small screens, so this will likely not be an issue on phones. 

I also think that even on mobile it might not be a problem to download a non-negligible amount of data. I don't see many people trying scientific computing from a low power machine with a slow connection.

PR #151, opened 2024-03-19
Add flake8 as a pre-commit

Carreau:
This should ensure that we maintain some quality to the Python code.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,5 @@
+[flake8]
+# W503: line break before binary operator
+ignore = W503
+max-line-length = 160
+max-complexity = 15
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,4 +4,7 @@ repos:
   hooks:
   - id: prettier
     types_or: [css, javascript]
-
+- repo: https://github.com/PyCQA/flake8
+  rev: '7.0.0'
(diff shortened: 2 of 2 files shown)


.flake8
```diff
@@ -0,0 +1,5 @@
+[flake8]
+# W503: line break before binary operator
+ignore = W503
+max-line-length = 160
+max-complexity = 15
```
Carreau, line 5: There is at least 1 line with 15 complexity.

steppi: I think some of this is redundant since CI already enforces conformance to black

https://github.com/jupyterlite/jupyterlite-sphinx/blob/36f3f6ed65d34b340554bd92fd3087e0f3bfd3e6/.github/workflows/main.yml#L39-L40

`max-line-length` is already 88. I'm not sure how `W503` interacts with black.

steppi: maybe have black fix the code in the pre-commit hooks? and just use `max-complexity` from flake8?

jtpio:
How about using Ruff directly otherwise? Since it's now used in many Jupyter projects.

Carreau:
I'm also happy with ruff, I just have flake 8 as an editor linter so I saw flake8 errors :-P

Carreau:
Closing as we have ruff now.

PR #150, opened 2024-03-19
Set lower default verbosity and add config options.

Carreau:
I think this should cover most of the cases.

See #149

PR #152, opened 2024-03-20
fix: Correct syntax in jupyter lite command list

matthewfeickert:
Amends PR #150 

* Add missing comma between commands.
* Removes dangling '--' that does not connect additional extra commands.

These changes unbreak https://github.com/scikit-hep/pyhf/pull/2458, and so should be minimally sufficient, though additional tests should be applied.

Tagging @steppi and @Carreau for review.


steppi:
Sorry it seems one cannot actually pass doit options to jupyterlite build like this and have them take affect. The output was only being suppressed because the command was failing. 0.13.0 will need to be yanked and further work will be needed to suppress the output.

matthewfeickert:
Closing in favor of PR #153. Thanks @steppi!

PR #153, opened 2024-03-20
Give a pragmatic solution to suppressing noisy output

steppi:
This PR replaces the broken means of suppressing noisy output from #150 with a pragmatic hack. (Note that although @Carreau submitted #150, I was the one who came up with this broken solution, and he was only implementing my idea).

The pragmatic hack I've made is to offer an option `jupyterlite_silence` in `config.py`. Setting this to `True` causes the `stderr` and `stout` from the call to `jupyterlite build` to be directed to `dev/null`. I've verified that this works for suppressing output, and that the notebooks still work. I have not made `jupyterlite_silence=True` the default because it's a drastic step which suppresses even warnings and errors.

I think it's reasonable to go with this hack, at least for the time being, until the necessary changes can be made in `jupyterlite`  removing unguarded `print` statements and making `doit` for tasks configurable.

Many apologies for jumping the gun and making a broken release. I will be sure to learn my lesson from this. I will need help revoking the bad release from PyPI.  

cc @Carreau, @martinRenou, @jtpio @matthewfeickert

### DIFF
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -641,7 +641,7 @@ def setup(app):
     app.add_config_value("jupyterlite_dir", str(app.srcdir), rebuild="html")
     app.add_config_value("jupyterlite_contents", None, rebuild="html")
     app.add_config_value("jupyterlite_bind_ipynb_suffix", True, rebuild="html")
-    app.add_config_value("jupyterlite_silence", False, rebuild=True)
+    app.add_config_value("jupyterlite_silence", True, rebuild=True)
 
     app.add_config_value("global_enable_try_examples", default=False, rebuild=True)
     app.add_config_value("try_examples_global_theme", default=None, rebuild=True)
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -548,11 +548,6 @@ def jupyterlite_build(app: Sphinx, error):
 
         jupyterlite_dir = str(app.env.config.jupyterlite_dir)
 
-        jupyterlite_debug = app.env.config.jupyterlite_debug
-        jupyterlite_log_level = app.env.config.jupyterlite_log_level
-        jupyterlite_verbosity = app.env.config.jupyterlite_verbosity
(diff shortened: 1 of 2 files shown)


matthewfeickert:
Thanks very much for being so fast and responsive on this @steppi! Very much appreciated. :pray: 

> Many apologies for jumping the gun and making a broken release. I will be sure to learn my lesson from this.

I think we've all been there before, so no worries. :)

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -580,37 +575,30 @@ def jupyterlite_build(app: Sphinx, error):
...
         assert all(
             [isinstance(s, str) for s in command]
         ), f"Expected all commands arguments to be a str, got {command}"
 
-        subprocess.run(command, cwd=app.srcdir, check=True)
+        kwargs = {"cwd": app.srcdir, "check": True}
+        if app.env.config.jupyterlite_silence:
+            kwargs["stdout"] = subprocess.DEVNULL
+            kwargs["stderr"] = subprocess.DEVNULL
+        subprocess.run(command, **kwargs)
```
Carreau, line 601: I think we should check the recur code of subprocess run (`CompletedProcess(args=..., returncode=1=...)`) and fail if it is Non-zero (i'll check). 

I also think that we might want to have stdout/err be PIPE and print them anyway _if_ the return code is non-zero

steppi: Good idea.

martinRenou:
> Many apologies for jumping the gun and making a broken release

No worries! I stopped counting the number of broken releases I made 🙈

steppi:
I've made it so `stdout` and `stderr` of `jupyterlite build` is still printed if the command fails following @Carreau's suggestion. I've also made `jupyterlite_silence = True` the default, following the suggestion of @rgommers here: https://github.com/scipy/scipy/issues/20277#issuecomment-2009794310.

steppi:
@Carreau, @martinRenou, @jtpio. Can we merge this and make a release? It would fix a broken build in one of SciPy’s CI jobs.

Carreau:
Ok, let's merge this and try. It's ok to do broken releases, it happens.

steppi:
Thanks @Carreau!

jtpio:
Yes, thanks all for working on this!

PR #155, opened 2024-03-22
Minor refactor + typing info

Carreau:
This adds mypy configuration to ignore voici (untyped), and slightly refactor to pass some kwargs explicitely.

It also pleases mypy.

### DIFF
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -4,12 +4,14 @@ from uuid import uuid4
 import shutil
 import glob
 import re
+from typing import Dict, Any
 
 from pathlib import Path
 
 from urllib.parse import quote
 
 import subprocess
+from subprocess import CompletedProcess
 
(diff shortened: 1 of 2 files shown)


agriyakhetarpal (approved):
Hi, @Carreau! I'm just helping @steppi a bit with code review on some of the pending PRs and I had a look at the changes here, and they look good on my end. Question: does adding `from __future__ import annotations` to the top of the file make sense too (and maybe with an accompanied [isort](https://docs.astral.sh/ruff/rules/missing-required-import/) rule as well after #156), considering it will be mandated with future Python versions someday? I don't think we're using anything complicated here that will break later.

agriyakhetarpal:
Thanks for the info, TIL that future-annotations can get us invalid types on earlier Python versions like that. I guess we would have to enable the `isort` rule before `pre-commit` starts to apply it across the codebase, but I suppose that's something to do in a separate PR.

Carreau:
No problem, thanks for the review and merge.

PR #161, opened 2024-03-29
Allow usage of global configuration values for `TryExamples` directive if provided by user

### LINKED ISSUE #157: `try_examples_global_button_text` does not propagate to custom `.. try_examples::` directives
## Description

When using the `try_examples_global_button_text` configuration value to customise the text displayed on every button, the text is used on buttons for docstring-based examples present in the API documentation for a library. However, buttons that are manually created/enabled using the `.. try_examples::` directive on code snippets outside of docstrings (such as those written in-line in `.rst` files, rather than `.py` files) do not respect this configuration and use the placeholder text "Try it with Jupyterlite` instead of the `try_examples_global_button_text` value set in `conf.py`.

It is to be noted that a [`:button_text`](https://jupyterlite-sphinx.readthedocs.io/en/latest/directives/try_examples.html#configuration) option is available in order to customise the text for the granularity of a specific button, but doing this for documentation websites where such buttons might be several in number might not make sense.

## Reproduce

In a standard documentation deployment provided with Sphinx, please follow these steps to reproduce:

1. Add `try_examples_global_button_text = "Your text here"` to `conf.py`
2. Add any code snippet to a `.rst` file tracked by Sphinx for inclusion in the built docs, like this:

```rst

The NumPy library provides several functions and tooling for array computing.
For example, you may try ``numpy.eye`` to represent an array that looks like an identity matrix:

.. try_examples::

  >>> import numpy as np
  >>> np.eye(2)
  array([[1, 0],
       [0, 1]])
```

3. Build the documentation, either locally, or in CI, or through any other methods.

A simple reproducer is available on a PR preview on Read the Docs for the PyWavelets documentation, as follows:

The example mentioned in the `pywt.families()` docstring has the button with the "Try it in your browser" text, as intended:

PyWavelets documentation PR preview for PR 728, intended

However, on the same page, code snippets coming from `wavelets.rst`, i.e., those not from `.. autofunction::` directives:

PyWavelets documentation PR preview for PR 728, unintended

don't have the correct text as mentioned above.

Here is a link to the specified page: https://pywavelets--728.org.readthedocs.build/en/728/ref/wavelets.html

xref: https://github.com/PyWavelets/pywt/pull/728

## Expected behavior

The expected behaviour here, IMO, is that `.. try_examples::`, in the absence of an `:button_text:` option, should use the value added to `try_examples_global_button_text` instead of the "Try it with Jupyterlite" boilerplate/placeholder text (it is a _global_ value, after all, as noted in its name!)

Therefore, a summary is as follows:

1. If `try_examples_global_button_text` is not `None`, apply it for all examples and `.. try_examples::` directives
2. If a `.. try_examples::` directive placed somewhere has a custom `:button_text:`, use that instead, overriding `try_examples_global_button_text`

## Context

- JupyterLite version: `jupyterlite-core` (`0.2.3`), `jupyterlite-sphinx` (`0.12.0`)
- Operating System and version: macOS Sonoma v14.3
- Browser and version: Brave Browser, version 1.64.109 with Chromium: 123.0.6312.58 (Official Build) (arm64)

agriyakhetarpal:
## Description

This PR updates the `TryExamplesDirective` class to use the configuration value in `conf.py` if specified, and otherwise use the "Try it with JupyterLite" text. This way, buttons that are created manually with the

```rst
.. try_examples::
```

directive in Sphinx will receive the same configuration value, and adding a `:button_text` input across several `.. try_examples::` directives to customise the button text will now **not** be required. The `try_examples_global_button_text` string will be used by default, instead of the current `"Try it with JupyterLite!"` text.

To explain this behaviour, an additional sentence has been added to the documentation for the `TryExamples` directive page.

Closes #157 

## Footnotes

This is essentially a breaking change of sorts for dependents who are not using a custom `:button_text` input already, since it will now change the text to what they have set in `try_examples_global_button_text` (in case they have set that). But it's surely not a very big change and I would consider it as more of an enhancement (cc: @steppi)

### DIFF
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -117,7 +117,7 @@ positioning of the button. The css for the example above is
 
 
 The `try_examples` directive has options
-* `:height:` To set a specific value for the height of the [iframe](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) containing the embeddednotebook.
+* `:height:` To set a specific value for the height of the [iframe](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) containing the embedded notebook.
 * `:button_text` To customize the text of the button that replaces the rendered examples with an embedded notebook.
 * `:theme:` This works the same as for the other JupyterLite-Sphinx directives.
 * `:example_class:` An html class to attach to the outer container for the rendered
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -205,9 +205,8 @@ allowing for specification of examples sections which should not be made interac
 
 The button text, theme, and warning text can be set globally with the config variables
 `try_examples_global_button_text`, `try_examples_global_theme`, and `try_examples_global_warning_text` in `conf.py`;
-and these apply both to automatically inserted directives and to manually inserted ones. The
-priority of these global settings is lower than that of the specific options set for a manually
-inserted directive, which will override the global settings on a per-directive basis.
+these apply both to automatically and manually inserted directives. Options set explicitly in a directive will
+override the global configuration.
 
 ```python
 global_enable_try_examples = True
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -118,13 +118,14 @@ positioning of the button. The css for the example above is
 
 The `try_examples` directive has these options:
 
-* `:height:` To set a specific value for the height of the [iframe](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) containing the embeddednotebook.
+* `:height:` To set a specific value for the height of the [iframe](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) containing the embedded notebook.
 * `:button_text` To customize the text of the button that replaces the rendered examples with an embedded notebook. The default value is `"Try it with JupyterLite!"`, or the value of the global configuration variable `try_examples_global_button_text` if it is set in `conf.py`.
 * `:theme:` This works the same as for the other JupyterLite-Sphinx directives.
 * `:example_class:` An html class to attach to the outer container for the rendered
 examples content and embedded notebook. This can be used in a custom css file to allow
 for more precise customization, eg. different button styles across different examples.
-* `:warning_text:` Prepend a markdown cell to the notebook containing this text, styled to make it clear this is intended as a warning.
+The default value is the value of the global configuration variable `try_examples_global_theme`, if it is set in `conf.py`.
+* `:warning_text:` Prepend a markdown cell to the notebook containing this text, styled to make it clear this is intended as a warning. The default value is the value of the global configuration variable `try_examples_global_warning_text` if it is set in `conf.py`, otherwise no warning text is added.
 
 Here's an example with some options set
 
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -19,7 +19,7 @@ Examples
 .. try_examples::
 
     Doctest examples sections are parsed and converted to notebooks. Blocks of text
-    like this become Markdown cells. Code blocks begin with ``>>>``. Contiguous blocks
+    like this become markdown cells. Codeblocks begin with ``>>>``. Contiguous blocks
     of code are combined into a single code cell.
 
     >>> x = 2
@@ -116,16 +116,14 @@ positioning of the button. The css for the example above is
 ```
 
 
-The `try_examples` directive has these options:
-
-* `:height:` To set a specific value for the height of the [iframe](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) containing the embedded notebook.
-* `:button_text` To customize the text of the button that replaces the rendered examples with an embedded notebook. The default value is `"Try it with JupyterLite!"`, or the value of the global configuration variable `try_examples_global_button_text` if it is set in `conf.py`.
+The `try_examples` directive has options
+* `:height:` To set a specific value for the height of the [iframe](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) containing the embeddednotebook.
+* `:button_text` To customize the text of the button that replaces the rendered examples with an embedded notebook.
 * `:theme:` This works the same as for the other JupyterLite-Sphinx directives.
 * `:example_class:` An html class to attach to the outer container for the rendered
 examples content and embedded notebook. This can be used in a custom css file to allow
 for more precise customization, eg. different button styles across different examples.
-The default value is the value of the global configuration variable `try_examples_global_theme`, if it is set in `conf.py`.
-* `:warning_text:` Prepend a markdown cell to the notebook containing this text, styled to make it clear this is intended as a warning. The default value is the value of the global configuration variable `try_examples_global_warning_text` if it is set in `conf.py`, otherwise no warning text is added.
+* `:warning_text:` Prepend a markdown cell to the notebook containing this text, styled to make it clear this is intended as a warning.
 
 Here's an example with some options set
 
@@ -206,7 +204,10 @@ allowing for specification of examples sections which should not be made interac
 
 
 The button text, theme, and warning text can be set globally with the config variables
-`try_examples_global_button_text`, `try_examples_global_theme`, and `try_examples_global_warning_text`.
+`try_examples_global_button_text`, `try_examples_global_theme`, and `try_examples_global_warning_text` in `conf.py`;
+and these apply both to automatically inserted directives and to manually inserted ones. The
+priority of these global settings is lower than that of the specific options set for a manually
+inserted directive, which will override the global settings on a per-directive basis.
 
 ```python
 global_enable_try_examples = True
@@ -215,7 +216,7 @@ try_examples_global_height = "200px"
 try_examples_global_warning_text = "Interactive examples are experimental and may not always work as expected."
 ```
 
-There is no option to set a global specific height because the proper height
+However, there is no option to set a global specific height because the proper height
 should depend on the size of the examples content. Again, the default height of
 the embedded notebook's iframe container matches the height of the associated
 rendered doctest example so that it takes up the same amount of space on the
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -19,7 +19,7 @@ Examples
 .. try_examples::
 
     Doctest examples sections are parsed and converted to notebooks. Blocks of text
-    like this become markdown cells. Codeblocks begin with ``>>>``. Contiguous blocks
+    like this become Markdown cells. Code blocks begin with ``>>>``. Contiguous blocks
     of code are combined into a single code cell.
 
     >>> x = 2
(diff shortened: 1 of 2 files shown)


agriyakhetarpal:
Happy to add a CHANGELOG entry for this if needed!

agriyakhetarpal:
Oops, let me run `black` and rebase

steppi:
> Happy to add a CHANGELOG entry for this if needed!

No need! The change log gets autogenerated by the prep release GitHub workflow.

steppi:
Thanks @agriyakhetarpal! Looking good so far. One thing: I think that all of the global config options should be propagated if a custom option isn't given, not just  `try_examples_globalbutton text`. You should also do `try_examples_global_theme`, and `try_examples_global_warning_text`.

agriyakhetarpal:
Right, I had thought of doing `try_examples_global_warning_text` too, but that would become a bigger breaking change, wouldn't it? People will have to override the global warning text and set it to `None` if they are using manual `TryExamples` directives somewhere. It does make sense, but I guess you're more aware of how downstream users would react, though. 

I agree about `try_examples_global_theme`, however – I'll make that change and for the other global config options.

agriyakhetarpal:
Fair point, didn't consider that the directive was/is still experimental – I have added in the requested changes! Now, one should be able to use the global warning text and use the global name for the theme/CSS container styling.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -387,10 +387,33 @@ def run(self):
...
         )
 
-        button_text = self.options.pop("button_text", "Try it with Jupyterlite!")
+        # Use button_text from try_examples_global_button_text if not provided
+        if self.env.config.try_examples_global_button_text is not None:
+            button_text = self.options.get(
+                "button_text", self.env.config.try_examples_global_button_text
+            )
+        else:
+            button_text = self.options.pop("button_text", "Try it with JupyterLite!")
```
steppi, line 396:
```suggestion
        default_button_text = self.env.config.try_examples_global_button_text
        if default_button_text  is None:
            default_button_text = "Try it with JupyterLite!"
        button_text = self.options.pop("button_text", default_button_text)
```

Two things. The use of `pop` instead of `get` is essential. We don't want "button_text" to remain in the `self.options` dictionary. The remaining key value pairs will passed as options into a url later on.

Second, I think the way I've suggested is cleaner because the control flow logic is used explicitly used to determine the default button text, and we avoid duplicating the logic to pop something from the dictionary.

agriyakhetarpal: Thanks! I had an issue come up where the button disappeared completely when I had used `pop` to display a global warning, I'm not sure why – so I had replaced everything with `get` instead.

I'm not sure why that happened, let me try to debug that.

agriyakhetarpal: Your logic was much better, and everything works as expected – tested all of them again. Addressed in 47621ff2dc743ca4491046bc97b9c04009e1291e.

docs/directives/try_examples.md
```diff
@@ -116,14 +116,16 @@ positioning of the button. The css for the example above is
...
+* `:button_text` To customize the text of the button that replaces the rendered examples with an embedded notebook. The default value is `"Try it with JupyterLite!"`, or the value of the global configuration variable `try_examples_global_button_text` if it is set in `conf.py`.
 * `:theme:` This works the same as for the other JupyterLite-Sphinx directives.
 * `:example_class:` An html class to attach to the outer container for the rendered
 examples content and embedded notebook. This can be used in a custom css file to allow
 for more precise customization, eg. different button styles across different examples.
-* `:warning_text:` Prepend a markdown cell to the notebook containing this text, styled to make it clear this is intended as a warning.
+The default value is the value of the global configuration variable `try_examples_global_theme`, if it is set in `conf.py`.
+* `:warning_text:` Prepend a markdown cell to the notebook containing this text, styled to make it clear this is intended as a warning. The default value is the value of the global configuration variable `try_examples_global_warning_text` if it is set in `conf.py`, otherwise no warning text is added.
```
steppi, line 128: I think it makes more sense for you not to change these, and then mention that `try_examples_global` config options apply both to automatically inserted `try_examples` directives, and manually inserted ones when the global configuration is discussed. For instance, I wouldn't consider the value of `try_examples_global_warning_text` to be a default, I think of it as setting to a non-default value in the global config. Also, I think it's better not to refer so heavily to the `try_examples_global` config options before they are introduced.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -387,10 +387,27 @@ def run(self):
...
+        default_button_text = self.env.config.try_examples_global_button_text
+        if default_button_text is None:
+            default_button_text = "Try it with JupyterLite!"
+        button_text = self.options.pop("button_text", default_button_text)
+
+        # Use warning_text from try_examples_global_warning_text if not provided
+        default_warning_text = self.env.config.try_examples_global_warning_text
+        if default_warning_text is None:
+            default_warning_text = "Interactive examples are experimental and may not always work as expected."
+        warning_text = self.options.pop("warning_text", default_warning_text)
```
steppi, line 400: The current default is for there to be no warning cell. I think we should keep that.

steppi (changes requested):
Thanks @agriyakhetarpal. This is looking better. A few minor suggestions.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -611,7 +634,7 @@ def jupyterlite_build(app: Sphinx, error):
         if completed_process.returncode != 0:
             if app.env.config.jupyterlite_silence:
                 print(
-                    "`jupyterlite build` failed but it's output has been silenced."
+                    "`jupyterlite build` failed but its output has been silenced."
```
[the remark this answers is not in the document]
agriyakhetarpal: Thank you for the advice. I will make a note of this in future PRs.

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: That makes sense – I reverted the change in 958b12e3697f747b54abaaf2b09990a8bf84ea4f. It should now pick up `self.env.config.try_examples_global_warning_text`, which can either be set to a string or it can be `None`. Let me know if you want a more stringent check.

docs/directives/try_examples.md
agriyakhetarpal: I agree with all of the three points. I have included these details under the "Global Configuration" section and reverted two of my previous commits in 7dad6e20120067617c2213eaa34515943038e096, and these details do make more sense there.

docs/directives/try_examples.md
```diff
@@ -213,7 +216,7 @@ try_examples_global_height = "200px"
 try_examples_global_warning_text = "Interactive examples are experimental and may not always work as expected."
 ```
 
-There is no option to set a global specific height because the proper height
```
steppi, line 216: I'm not sure this is an improvement. It's distance from a contrasting statement makes the "However" seem out of place to me.

agriyakhetarpal: Should we allow overriding the height – we don't do that currently?

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -387,10 +387,26 @@ def run(self):
...
+        # into the notebook
         height = self.options.pop("height", None)
-        example_class = self.options.pop("example_class", "")
-        warning_text = self.options.pop("warning_text", None)
+
+        # Use example_class from try_examples_global_theme if not provided
+        default_example_class = self.env.config.try_examples_global_theme
+        if default_example_class is None:
+            default_example_class = ""
+        example_class = self.options.pop("example_class", default_example_class)
```
steppi, line 409: I don't think we need a specific comment like this for each config option. I think it's clear enough to have a single comment above the processing of options that says something like:

Use global configuration values from `conf.py` in manually inserted directives if they are provided and the user has not specified a config value in the directive itself.

Also I think it makes sense to make `height` either before or after all of the other instances, since it is treated differently. You can mention here there is no global option for height.

docs/directives/try_examples.md
agriyakhetarpal: I got confused with something else, please don't mind. Since we do have a method to set a global minimum height with the JSON file, I think removing "However" is okay – it makes it feel like there is a setback when there is none.

agriyakhetarpal: Addressed this in 998cd260a78a8b337682116d8a87404b0f19cd0a

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: Thanks, I added a single comment in 62b3dea0b88a27e4802c11f388e8be6e77eaeab4 for the applicable options, and added another comment above the `height` configuration to signify the difference between the rest

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -387,10 +387,26 @@ def run(self):
...
+        default_warning_text = self.env.config.try_examples_global_warning_text
+        warning_text = self.options.pop("warning_text", default_warning_text)
+
+        default_example_class = self.env.config.try_examples_global_theme
+        if default_example_class is None:
+            default_example_class = ""
+        example_class = self.options.pop("example_class", default_example_class)
+
+        # No global option for height is available, so we use the one provided
+        # manually in the directive.
```
steppi, line 408: Not technically true because of the min height option in `try_examples.json`.

I think just "A global height cannot be set in conf.py" is clear enough

docs/directives/try_examples.md
```diff
@@ -204,7 +204,10 @@ allowing for specification of examples sections which should not be made interac
 
 
 The button text, theme, and warning text can be set globally with the config variables
-`try_examples_global_button_text`, `try_examples_global_theme`, and `try_examples_global_warning_text`.
+`try_examples_global_button_text`, `try_examples_global_theme`, and `try_examples_global_warning_text` in `conf.py`;
+and these apply both to automatically inserted directives and to manually inserted ones. The
+priority of these global settings is lower than that of the specific options set for a manually
+inserted directive, which will override the global settings on a per-directive basis.
```
steppi, line 210:
```suggestion
`try_examples_global_button_text`, `try_examples_global_theme`, and `try_examples_global_warning_text` in `conf.py`;
these apply both to automatically and manually inserted directives. Options set explicitly in a directive will
override the global configuration.
```

A semicolon should link two independent clauses which form a logical whole. The effect is to imply a pause of shorter duration than a hard stop. Starting the second clause with an "and" makes it a dependent clause. Also I think this can be expressed more concisely.

steppi (changes requested):
Thanks @agriyakhetarpal. this is in good shape now. Just a couple minor suggestions on language.

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: You're right. Addressed this in cd542e3c48ba2935881f13a8079f56e6c84a8009.

steppi (approved):
Thanks @agriyakhetarpal! This looks good now.

Issue #114, opened 2023-11-13
Add an `:app:` parameter to the `jupyterlite` directive


jtpio:
### Problem

Currently we have a couple of directives for either using the full JupyterLab interface, Notebook or REPL.

However this could be a bit tedious to maintain over time if more apps are added.

### Proposed Solution

As noticed in https://github.com/jupyterlite/jupyterlite-sphinx/pull/113#discussion_r1391094214, we could start introduce an `:app:` parameter to the directive.

To use Jupyter Notebook, this would mean using the `notebooks` app from JupyterLite:

```rst
.. jupyterlite::
   :width: 100%
   :height: 600px
   :app: notebooks
   :prompt: Try classic Notebook!
```

For JupyterLab:

```rst
.. jupyterlite::
   :width: 100%
   :height: 600px
   :app: lab
```

For REPL:

```rst
.. jupyterlite::
   :width: 100%
   :height: 600px
   :app: repl
```

### Additional context

Follow-up to https://github.com/jupyterlite/jupyterlite-sphinx/pull/113#discussion_r1391094214

PR #165, opened 2024-04-16
Add the option to open JupyterLite window in new tab

melissawm:
My use case is the following: I'd like to use the jupyterlite directive to open the **full page** I am visiting as a Jupytext notebook. This PR adds that option to the JupyterLite directive.  

Pending is finding the correct path for opening the current file __as an ipynb file__: since the source is an .md file, it will be converted into an .ipynb file at some point in the build process. For now, however, Sphinx can't process the directive since it will try to embed a link to the ipynb file before reading the current source .md file, and complain about the missing reference.

I am not sure if I am following the right path here so please let me know if I should approach this differently. 

cc @agriyakhetarpal

### DIFF
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -128,7 +128,7 @@ class _InTab(Element):
         self.lab_src = f'{prefix}/{app_path}{f"?{options}" if options else ""}'
 
         super().__init__(
-            "",
+            rawsource,
             **attributes,
         )
 
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -106,6 +106,40 @@ class _PromptedIframe(Element):
             f'width="{self["width"]}" height="{self["height"]}" class="jupyterlite_sphinx_raw_iframe"></iframe>'
         )
 
+class _InTab(Element):
+    def __init__(
+        self,
+        rawsource="",
+        *children,
+        prefix=JUPYTERLITE_DIR,
+        notebook=None,
+        lite_options={},
+        **attributes,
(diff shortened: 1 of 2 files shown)


melissawm:
Here is an example of this working (partially): https://melissawm.github.io/scipy/tutorial/stats/sampling.html

martinRenou:
Thanks for contributing! 

It's deployed in the readthedocs preview: https://jupyterlite-sphinx--165.org.readthedocs.build/en/165/directives/jupyterlite.html

Changes look good to me. I was about to suggest doing the same for other directives, but when we implement https://github.com/jupyterlite/jupyterlite-sphinx/issues/114, we can make sure the `:new_tab:` parameter works well with the coming `:app:` parameter.

Leaving this open to let other people have a look.

melissawm:
I'm happy to do a follow up for that in the future, I might need some time togo over the app parameter and understand the effects if that's alright. Thanks!

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -107,6 +107,39 @@ def html(self):
...
+
+        options = "&".join(
+            [f"{key}={quote(value)}" for key, value in lite_options.items()]
+        )
+        self.lab_src = f'{prefix}/{app_path}{f"?{options}" if options else ""}'
+
+        super().__init__(
+            "",
+            **attributes,
+        )
```
Carreau, line 133: Maybe `rawsource` here ?
```suggestion
        super().__init__(
            rawsource,
            **attributes,
        )
```

Carreau (approved):

Carreau (approved):
All good on my side, I'll leave ~24h  in case of objections and merge after that.

Thanks @melissawm

Carreau:
Completely forgot to merge, sorry. Merging.

PR #169, opened 2024-04-29
Pass additional configuration options to the `jupyter lite build` command

### LINKED ISSUE #168: more flexible ways to specify preinstalled packages
### Problem

as per ,https://jupyterlite-sphinx.readthedocs.io/en/latest/configuration.html#pre-installed-packages> 
one can specify the preinstalled packages in a file named `environment.yml`
i see 2 limitations:
- given that this filename is already heavily used, it might help if one could decide to store this information under a different filename, (like e.g. `jupyterlite.yml`)
- also, it would be really helpful to support for a `pip`-compliant format (like `requirements.txt`)

* my main use case is for writing courses, and in that sense what is usually referred to as `docs` is the main folder in my case
* in that context I use `environment.yml` already to store the "usual" - i.e. not lite - environment to use the material (typically jupyterlab and similar)

### Proposed Solution

to this end, I would propose to support a new configuration variable in `conf.py` named `jupyterlite_environment` or similar
that would use `conda` or `pip` depending on the file extension (`txt`→`pip` and `yml` or `yaml` → conda);

or alternatively a more elaborate scheme where e.g.
```python
jupyterlite_environment = {'pip': 'requirements-lite.txt'}
# or
jupyterlite_environment = {'conda': 'environment-lite.yml'}
```

I understand that the conda flavour is more powerful in that it contains the python version; however not everybody is fluent in conda, so maybe the `pip` flavour makes sense too

### Additional context

as far as I am concerned, supporting `pip` is less crucial than just being able to use another yaml file, so this could be implemented in 2 separate steps

agriyakhetarpal:
## Description

This PR closes #168. It aims to implement additional options from the JupyterLite CLI, i.e., the `jupyter lite build` command that is currently wrapped in a subprocess call inside a Sphinx app. This enables the use of, say, custom environments from a file as requested in https://github.com/jupyterlite/jupyterlite-sphinx/issues/168#issuecomment-2079472156, allowing overrides in case of conflicts where `environment.yml` might refer to a file that is not being used for `jupyterlite-sphinx` but for setting up general environments with `mamba` or `conda`.

## Changes made

- Added `jupyterlite_build_command_options` as a configuration option to allow power users of `jupyterlite-sphinx` to customise their JupyterLite build procedure with a dictionary of options.

## Task list

- [x] Implement ways to add options from https://jupyterlite.readthedocs.io/en/latest/reference/cli.html#common-parameters
- [x] Add documentation about these user-configurable options
- [ ] Some of the options are available to be overridden by environment variables, adjust the default values in the `setup()` function?
- [x] Address potential conflicts with options, as noted by @steppi in #168, copying here for ease of access:

> except we might want to document how to avoid having options that conflict with the ones that are currently included, and put in place safeguards to prevent conflicting options

## Additional context

Since the `jupyterlite-pyodide` kernel doesn't support pre-installing packages right now, it's not possible to add `requirements.txt` files. Users shall have to use the `emscripten-forge` channel and the `xeus` kernel to get the packages they need.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -29,9 +29,9 @@ You would need `jupyterlite-xeus` installed in your docs build environment.
 
 You can pre-install packages by adding an `environment.yml` file in the docs directory, with `xeus-python` defined as one of the dependencies. It will pre-build the environment when running the `jupyter lite build`.
 
-Furthermore, this automatically installs any labextension that it finds, for example, installing `ipyleaflet` will make `ipyleaflet` work without the need to manually install the `jupyter-leaflet` labextension.
+Furthermore, this automatically installs any labextension that it founds, for example installing ipyleaflet will make ipyleaflet work without the need to manually install the jupyter-leaflet labextension.
 
-Say you want to install NumPy, Matplotlib and ipycanvas, it can be done by creating the `environment.yml` file with the following content:
+Say you want to install NumPy, Matplotlib and ipycanvas, it can be done by creating the environment.yml file with the following content:
 
 ```yaml
 name: xeus-python-kernel
@@ -56,7 +56,7 @@ jupyterlite_config = "jupyterlite_config.json"
 ## Disable the `.ipynb` docs source binding
 
 By default, jupyterlite-sphinx binds the `.ipynb` source suffix so that it renders Notebooks included in the doctree with JupyterLite.
-This is known to bring warnings with plugins like `sphinx-gallery`, or to conflict with `nbsphinx`.
+This is known to bring warnings with plugins like sphinx-gallery, or to conflict with nbsphinx.
 
 You can disable this behavior by setting the following config:
 
@@ -64,11 +64,11 @@ You can disable this behavior by setting the following config:
 jupyterlite_bind_ipynb_suffix = False
 ```
 
-### Suppressing JupyterLite logging
+### Suppressing jupyterlite logging
 
 `jupyterlite` can produce large amounts of output to the terminal when docs are building.
 By default, this output is silenced, but will still be printed if the invocation of
-`jupyter lite build` fails. To unsilence this output, set
+`jupyterlite build` fails. To unsilence this output, set
 
 ```python
 jupyterlite_silence = False
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -82,14 +82,13 @@ You can pass additional arguments to the `jupyter lite build` command by passing
 
 ```python
 jupyterlite_build_command_options = {
-    "port": 8888,
     "XeusAddon.environment_file": "jupyterlite_environment.yml",
-    "source-date-epoch": "315532800",
+    ...
     }
 ```
 
-This will launch the JupyterLite site on port 8888, configure the `jupyterlite-xeus` kernel with a custom
-`jupyterlite_environment.yml` file, and configure the source date epoch to ensure reproducible builds.
+This will launch the JupyterLite site, configuring the `jupyterlite-xeus` kernel with a custom
+`jupyterlite_environment.yml` file.
 
 :::{attention}
 The `--contents`, `--output-dir`, and `--lite-dir` options are not supported in this scenario, as they are set by
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -90,13 +90,9 @@ jupyterlite_build_command_options = {
 This will launch the JupyterLite site, configuring the `jupyterlite-xeus` kernel with a custom
 `jupyterlite_environment.yml` file.
 
-:::{attention}
 The `--contents`, `--output-dir`, and `--lite-dir` options are not supported in this scenario, as they are set by
 the [`jupyterlite_contents`](#jupyterlite-content) and the[`jupyterlite_dir`](#jupyterlite-dir) configuration
 options, respectively, as described above.
-:::
 
-:::{seealso}
 The full list of available options can be found by running the `jupyter lite build --help-all` command. For more
 information, please visit the [JupyterLite documentation](https://jupyterlite.readthedocs.io/en/stable/reference/cli.html#usage).
-:::
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -98,6 +98,7 @@ These can instead be set with
 the [`jupyterlite_contents`](#jupyterlite-content) and the[`jupyterlite_dir`](#jupyterlite-dir) configuration
 options described above.
 
-These options take precedence over the options provided by the user such as via environment variables
-or through JSON-based configuration files.
 This is an advanced feature and users are responsible for providing sensible command line options.
+The standard precedence between `jupyter lite build` CLI options and other means of configuration apply.
+See the [jupyter lite CLI](https://jupyterlite.readthedocs.io/en/latest/reference/cli.html) documentation
+for more info.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -97,3 +97,5 @@ The options `--contents`, `--output-dir`, and `--lite-dir` cannot be passed to `
 These can instead be set with
 the [`jupyterlite_contents`](#jupyterlite-content) and the[`jupyterlite_dir`](#jupyterlite-dir) configuration
 options described above.
+
+This is an advanced feature and users are responsible for providing sensible command line options.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -78,7 +78,7 @@ in your Sphinx `conf.py`.
 
 ## Additional CLI arguments for `jupyter lite build`
 
-You can pass additional arguments to the `jupyter lite build` command by passing a dictionary of arguments to the `jupyterlite_build_command_args` configuration option, without the hyphens prepended to the keys. Here's an example:
+You can pass additional arguments to the `jupyter lite build` command by passing a dictionary of arguments to the `jupyterlite_build_command_options` configuration option, without the hyphens prepended to the keys. Here's an example:
 
 ```python
 jupyterlite_build_command_options = {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -78,21 +78,22 @@ in your Sphinx `conf.py`.
 
 ## Additional CLI arguments for `jupyter lite build`
 
-You can pass additional arguments to the `jupyter lite build` command by passing a dictionary of arguments to the `jupyterlite_build_command_options` configuration option, without the hyphens prepended to the keys. Here's an example:
+Additional arguments can be passed to the `jupyter lite build` command using the configuration
+option `jupyterlite_build_command_options` in `conf.py`. The following example shows how to
+specify an alternative location for the `xeus` kernel's `environment.yml` file as discussed
+[here](https://github.com/jupyterlite/xeus#usage).
 
 ```python
 jupyterlite_build_command_options = {
     "XeusAddon.environment_file": "jupyterlite_environment.yml",
-    ...
     }
 ```
 
-This will launch the JupyterLite site, configuring the `jupyterlite-xeus` kernel with a custom
-`jupyterlite_environment.yml` file.
+This causes the additional option `--XeusAddon.environment_file=jupyterlite_environment.yml`
+to be passed to `jupyter lite build` internally within `jupyterlite-sphinx`. Note that one
+does not include the leading dashes, `--`, in the keys.
 
-The `--contents`, `--output-dir`, and `--lite-dir` options are not supported in this scenario, as they are set by
+The options `--contents`, `--output-dir`, and `--lite-dir` cannot be passed to `jupyter lite build` in this way.
+These can instead be set with
 the [`jupyterlite_contents`](#jupyterlite-content) and the[`jupyterlite_dir`](#jupyterlite-dir) configuration
-options, respectively, as described above.
-
-The full list of available options can be found by running the `jupyter lite build --help-all` command. For more
-information, please visit the [JupyterLite documentation](https://jupyterlite.readthedocs.io/en/stable/reference/cli.html#usage).
+options described above.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -29,9 +29,9 @@ You would need `jupyterlite-xeus` installed in your docs build environment.
 
 You can pre-install packages by adding an `environment.yml` file in the docs directory, with `xeus-python` defined as one of the dependencies. It will pre-build the environment when running the `jupyter lite build`.
 
-Furthermore, this automatically installs any labextension that it founds, for example installing ipyleaflet will make ipyleaflet work without the need to manually install the jupyter-leaflet labextension.
+Furthermore, this automatically installs any labextension that it finds, for example, installing `ipyleaflet` will make `ipyleaflet` work without the need to manually install the `jupyter-leaflet` labextension.
 
-Say you want to install NumPy, Matplotlib and ipycanvas, it can be done by creating the environment.yml file with the following content:
+Say you want to install NumPy, Matplotlib and ipycanvas, it can be done by creating the `environment.yml` file with the following content:
 
 ```yaml
 name: xeus-python-kernel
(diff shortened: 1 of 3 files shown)


steppi:
Thanks @agriyakhetarpal. Is there any reason why we can't just let the users pass a dictionary of options with values in `conf.py`, and check that none of the options they've provided conflict with the existing ones? To me, this seems preferable to manually adding new options every time we discover something that would make sense to have.

agriyakhetarpal:
> Thanks @agriyakhetarpal. Is there any reason why we can't just let the users pass a dictionary of options with values in `conf.py`, and check that none of the options they've provided conflict with the existing ones? To me, this seems preferable to manually adding new options every time we discover something that would make sense to have.

This is what I was thinking as well with allowing a custom `jupyter lite build` command – a dictionary would indeed serve that purpose better, sure. Do you have something in mind about checking for conflicts with existing options, i.e., do we plan to check against the current trio of `--contents`, `--output-dir`, and `--lite-dir` – and gracefully raise an error, and for _additional_ options that we currently don't have but plan to add with this PR, let the `jupyter lite` CLI raise errors on conflicts on its own accord?

steppi:
> Do you have something in mind about checking for conflicts with existing options, i.e., do we plan to check against the current trio of `--contents`, `--output-dir`, and `--lite-dir` – and gracefully raise an error, and for _additional_ options that we currently don't have but plan to add with this PR, let the `jupyter lite` CLI raise errors on conflicts on its own accord?

Yeah, checking for conflicts with the existing options, and if found raising, but leaving anything beyond that to the  `jupyter lite` CLI sounds like the right approach. When we document this, we can show what the default command is and let users know there will be an error if they supply options that conflict with that, but beyond that, they are responsible for entering sensible options.

agriyakhetarpal:
Thanks, here's a very crude implementation in a0328121625c909fa2333f1ff4dd53118d06c134, I feel that `snake_case` is better than `kebab-case`, especially inside Python files like it is so here. I think it needs further refinement, and additions to the documentation of course. Open to suggestions.

agriyakhetarpal:
Thanks, here's a very crude implementation in 6eb6680664cb5096fba50d1a7e446409533197cd – I feel that `snake_case` is better than `kebab-case`, especially inside Python files like it is so here. I think it needs further refinement, and additions to the documentation of course. Open to suggestions.

Also, I un-silenced JupyterLite in 617a614831fd850b3462a320bd7d09f911bb413f, maybe it's good for us because we can inspect the logs both when checks pass and when they don't? Happy to revert the change if you disagree!

steppi:
Thanks @agriyakhetarpal. This is starting to look good. I'll take a closer look later today.

> Also, I un-silenced JupyterLite in [617a614](https://github.com/jupyterlite/jupyterlite-sphinx/commit/617a614831fd850b3462a320bd7d09f911bb413f), maybe it's good for us because we can inspect the logs both when checks pass and when they don't? Happy to revert the change if you disagree!

I see, what you've done is not to un-silence the output by default, but just override the default in the `conf.py` for `jupyterlite-sphinx`'s docs. We might want to do that, but if so it should go into a separate PR and be discussed separately.

Can you add the example from https://github.com/jupyterlite/jupyterlite-sphinx/issues/168#issuecomment-2079472156 for changing the path to where `xeus` looks for the environment file to the docs?

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -614,6 +618,22 @@ def jupyterlite_build(app: Sphinx, error):
...
+                if key in ["contents", "output_dir", "lite_dir"]:
+                    jupyterlite_command_error_message = """
+                Cannot use [contents, output_dir, lite_dir] in jupyterlite_build_command_options.
+                Please use the corresponding option directly in conf.py.
+                For a list of available options, run the 'jupyter lite build --help-all'
+                command and refer to the documentation for jupyterlite-sphinx:
+                https://jupyterlite-sphinx.readthedocs.io/en/stable/configuration.html
+                """
+                    raise ValueError(jupyterlite_command_error_message.format(key=key))
+                command.extend([f"--{key.replace('_', '-')}", value])
```
steppi, line 635: I think it's an unnecessary complication to try to let users pass in commands with snake case when the option flag is not snake case. Also, what about commands that really have underscores like

```
jupyter lite build --XeusAddon.environment_file=some_path/to/environment.yaml
```?

**Comment by agriyakhetarpal** [comment]
> I see, what you've done is not to un-silence the output by default, but just override the default in the conf.py for jupyterlite-sphinx's docs. We might want to do that, but if so it should go into a separate PR and be discussed separately.

Ah, apologies if I was too terse. I indeed mean to do it for `jupyterlite-sphinx`, i.e., here, and not for other repositories. However, I agree that it should go into another PR, not here – I'll revert the commit.

File: jupyterlite_sphinx/jupyterlite_sphinx.py (hunk 1 above)
Comment (agriyakhetarpal) [reply jupyterlite_sphinx/jupyterlite_sphinx.py]: You're right, it could definitely break things that way – we can ask people to type `kebab-case`, and silently changing to snake_case isn't ideal for user behaviour. Addressed this in 9f13354a53cc8d9b118e7ae595e7b535a354ba65.

File: docs/configuration.md
Comment (steppi) [inline docs/configuration.md:1]: If you want to clean up the formatting that would be better to do in a separate PR.

Comment (agriyakhetarpal) [reply docs/configuration.md:1]: Reverted this in 511b5c6e7c692371d33c02496ce5703a24416fd7. I'll make a new PR with these changes later!

**Comment by agriyakhetarpal** [comment]
> Can you add the example from https://github.com/jupyterlite/jupyterlite-sphinx/issues/168#issuecomment-2079472156 for changing the path to where xeus looks for the environment file to the docs?

I have now done this in f6e6d3b2154fe42b22528d3bb1b108ad04d3ab84, happy to receive comments on the wording and if the location of this section is alright. I enabled some admonitions for this page by enabling the `colon_fence` MyST extension in 0b5e4996648f7363774d2f10b531a6d020ee856f, please let me know if that is out of scope for this PR.

File: docs/configuration.md [hunk 2]
```diff
@@ -75,3 +75,29 @@ jupyterlite_silence = False
...
+
+You can pass additional arguments to the `jupyter lite build` command by passing a dictionary of arguments to the `jupyterlite_build_command_options` configuration option, without the hyphens prepended to the keys. Here's an example:
+
+```python
+jupyterlite_build_command_options = {
+    "port": 8888,
+    "XeusAddon.environment_file": "jupyterlite_environment.yml",
+    "source-date-epoch": "315532800",
+    }
+```
```
steppi, line 89: Have you tested this out. I got the error

```
Extension error (jupyterlite_sphinx.jupyterlite_sphinx):
Handler <function jupyterlite_build at 0x717566aa3010> for event 'build-finished' threw an exception (exception: Expected all commands arguments to be a str, got ['jupyter', 'lite', 'build', '--debug', '--contents', '/home/birbir/interactive_doc_experiments/doc/source/_contents', '--output-dir', '/home/birbir/interactive_doc_experiments/doc/build/html/lite', '--apps', 'notebooks', '--apps', 'edit', '--apps', 'lab', '--apps', 'repl', '--apps', 'tree', '--apps', 'consoles', '--apps', 'voici', '--lite-dir', '/home/birbir/interactive_doc_experiments/doc/source', '--port', 8888])
make: *** [Makefile:25: html] Error 2
```

because port is an `int` here, but `subprocess.run` expects all command args to be strings. If you want to be able to allow integers here, they need to be converted to strings.

Also, in this context, I don't think it makes any sense to specify a `port` here, so let's not use that in the example. It's lucky you included it though, because we might have missed the bug.

I think just having the `"XeussAddon.environment_file"` option is file here.

docs/configuration.md
```diff
@@ -75,3 +75,29 @@ jupyterlite_silence = False
...
+:::{attention}
+The `--contents`, `--output-dir`, and `--lite-dir` options are not supported in this scenario, as they are set by
+the [`jupyterlite_contents`](#jupyterlite-content) and the[`jupyterlite_dir`](#jupyterlite-dir) configuration
+options, respectively, as described above.
+:::
+
+:::{seealso}
+The full list of available options can be found by running the `jupyter lite build --help-all` command. For more
+information, please visit the [JupyterLite documentation](https://jupyterlite.readthedocs.io/en/stable/reference/cli.html#usage).
+:::
```
steppi, line 103: I think all of these admonitions are drawing too much attention to what's supposed to be an advanced feature that will only be used by people who know what they're doing.

docs/configuration.md
agriyakhetarpal: Addressed in d1acb8874c1d39f0bb70e13ca0a9b7da368492ee, I have now just kept the `"XeussAddon.environment_file"` option and clarified the sentence below it

docs/configuration.md
agriyakhetarpal: That makes sense, removed them in 6cc2cabe7ca868045f6a45707897e24b5f8f39e7 (we can look into adding the `colon_fence` plugin later)

docs/configuration.md
```diff
@@ -75,3 +75,24 @@ jupyterlite_silence = False
 ```
 
 in your Sphinx `conf.py`.
+
```
steppi, line 78: What do you think of this wording?

Additional arguments can be passed to the `jupyter lite build` command using the configuration
option `jupyterlite_build_command_options` in `conf.py`. The following example shows how to
specify an alternative location for the `xeus` kernel's `environment.yml` file as discussed
[here](https://github.com/jupyterlite/xeus?tab=readme-ov-file#usage).

```python
jupyterlite_build_command_options = {
    "XeusAddon.environment_file": "jupyterlite_environment.yml",
    }
```

This causes the additional option `--XeusAddon.environment_file=jupyterlite_environment.yml`
to be passed to `jupyter lite build` internally within `jupyterlite-sphinx`. Note that one
does not include the leading dashes, `--`, in the keys.

The options `--contents`, `--output-dir`, and `--lite-dir` cannot be passed to `jupyter lite build` in this way. These can instead be set with
the [`jupyterlite_contents`](#jupyterlite-content) and [`jupyterlite_dir`](#jupyterlite-dir) configuration options described above.

agriyakhetarpal: This reads well to me. Thanks! Committing this change.

steppi: and then I don't think the thing about the full list of commands should be included. It doesn't actually include the full list of commands, and the most useful ones will probably be for `Addons` which aren't included there.

agriyakhetarpal: Should we point to some particular JupyterLite Addons just in case, while not being too verbose about it?

steppi: I don't think so. I don't think we want to explicitly encourage people to use this, but it's good it's there in case someone needs it. I expect anyone who needs this level of customization will be able to figure things out from the info here.

agriyakhetarpal: Thanks, addressed in 7b954c44f9724b8279c3ca0a6146cd3c70bc26f4.

steppi: Maybe add a sentence like,

```
This is an advanced features and users are responsible for providing sensible command line options.
```

agriyakhetarpal: Added the sentence in d49069ef7ab65494640c72ee2bcf578547524a6c.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -572,6 +572,10 @@ def jupyterlite_build(app: Sphinx, error):
 
         jupyterlite_dir = str(app.env.config.jupyterlite_dir)
 
+        jupyterlite_build_command_options: Dict[str, Any] = (
+            app.env.config.jupyterlite_build_command_options
+        )
```
agriyakhetarpal, line 577: Should we keep this as `[str, str]` since we're not pre-processing any of the values from the dict?
```suggestion
        jupyterlite_build_command_options: Dict[str, str] = (
            app.env.config.jupyterlite_build_command_options
        )
```

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -614,6 +618,22 @@ def jupyterlite_build(app: Sphinx, error):
...
+                if key in ["contents", "output_dir", "lite_dir"]:
+                    jupyterlite_command_error_message = """
+                Cannot use [contents, output_dir, lite_dir] in jupyterlite_build_command_options.
+                Please use the corresponding option directly in conf.py.
+                For a list of available options, run the 'jupyter lite build --help-all'
+                command and refer to the documentation for jupyterlite-sphinx:
+                https://jupyterlite-sphinx.readthedocs.io/en/stable/configuration.html
+                """
+                    raise ValueError(jupyterlite_command_error_message)
+                command.extend([f"--{key}", value])
```
steppi, line 635:
```suggestion
                    raise ValueError(jupyterlite_command_error_message)
                command.extend([f"--{key}", str(value)])
```

I think it's fine just to convert these to `str` here, and then you can leave the `Any` in the type hints.

steppi (changes requested):
A couple more suggestions for the exception raised when conflicting options are supplied.

steppi:
@agriyakhetarpal. Do you still consider this to be a work in progress, and if so, what else are you planning to do?

agriyakhetarpal:
I just merged `main` and I was editing the PR description just now :) I think this should be good to go very soon.

agriyakhetarpal:
Edited the title and the description. I think this is the only remaining point that we can think to address:

> Some of the options are available to be overridden by environment variables, adjust the default values in the `setup()` function?

If someone does not pass in a custom option via `conf.py` through this configuration value, they can alternatively do so using environment variables. However, it is possible that options set via environment variables can cause conflicts with those  set via `conf.py` and it would be difficult for the user to debug that? I don't see anything in the CLI documentation about whether the environment variable takes priority or whether the option in the command does (it is the latter, I just confirmed, and it's usually so based on how most CLIs work – but should we document that, considering we would like to reduce irrelevant bug reports?)

steppi:
> but should we document that, considering we would like to reduce irrelevant bug reports?)

Yes. I think it would be good to add a sentence saying that options in the `jupyter lite build` command take precedence over options in the config file. Before the sentence about this being an advanced feature I think.

steppi:
> Yes. I think it would be good to add a sentence saying that options in the `jupyter lite build` command take precedence over options in the config file. Before the sentence about this being an advanced feature I think.

Or I guess there are environment variables, a json config file, and command line options?

agriyakhetarpal:
> Or I guess there are environment variables, a json config file, and command line options?

The documentation seems to be quite sparse about it – the JSON configuration file isn't mentioned on the CLI documentation page, and https://jupyterlite.readthedocs.io/en/stable/howto/configure/config_files.html explains which JSON files affect the build command, but there is no reference about the priority as such. I think the priority is CLI > environment variables > JSON configuration file, but I need to check for the second and third comparisons.

agriyakhetarpal:
Or, we can not state the comparison directly through the sentence added in a9ca8867f44bd56bfa4a82c206e947bf2dedcb25, considering this is a feature for power users and they can figure out things as they try it (and no one would, in realistic contexts, use a mixture of all three: `conf.py`, environment variables, and a configuration JSON file altogether – a combination of the first and the third options is more likely).

This is ready for re-review/merging on my end.

docs/configuration.md
```diff
@@ -75,3 +75,29 @@ jupyterlite_silence = False
...
+does not include the leading dashes, `--`, in the keys.
+
+The options `--contents`, `--output-dir`, and `--lite-dir` cannot be passed to `jupyter lite build` in this way.
+These can instead be set with
+the [`jupyterlite_contents`](#jupyterlite-content) and the[`jupyterlite_dir`](#jupyterlite-dir) configuration
+options described above.
+
+These options take precedence over the options provided by the user such as via environment variables
+or through JSON-based configuration files.
+This is an advanced feature and users are responsible for providing sensible command line options.
```
steppi, line 103:
```suggestion
This is an advanced feature and users are responsible for providing sensible command line options.
The standard precedence between `jupyter lite build` CLI options and other means of configuration apply.
See the [jupyter lite CLI](https://jupyterlite.readthedocs.io/en/latest/reference/cli.html) documentation
for more info.
```

steppi (approved):

steppi:
Thanks @agriyakhetarpal, this looks good to me now.

PR #173, opened 2024-05-01
Hide buttons generated by the `TryExamples` directive by default

agriyakhetarpal:
## Description

This PR hides the buttons from the `TryExamples` directive by default. This is helpful because the current method of ignoring particular pages based on their URL first loads the buttons and then hides them, which means that they are briefly displayed on that page in the documentation and then hidden – keeping them hidden and then selectively enabling them will minimise this latency and improve user experience.

## Changes made

1. Made all `try_examples_button` buttons `hidden` at the time of generation by adding this keyword to the `classList`
2. Updated the `loadTryExamplesConfig` function in `jupyterlite_sphinx.js` to check for the `try_examples_button hidden` class instead and then remove `hidden` from the `classList`
3. 

## Additional context

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ docs/build
 build
 dist
 .jupyterlite.doit.db
-/.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ docs/build
 build
 dist
 .jupyterlite.doit.db
+/.vscode/
--- a/jupyterlite_sphinx/jupyterlite_sphinx.css
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.css
@@ -58,6 +58,7 @@
   background:
     radial-gradient(farthest-side, #ffa516 94%, #0000) top/8px 8px no-repeat,
     conic-gradient(#0000 30%, #ffa516);
+  mask: radial-gradient(farthest-side, #0000 calc(100% - 8px), #000 0);
   -webkit-mask: radial-gradient(farthest-side, #0000 calc(100% - 8px), #000 0);
   animation: l13 1s infinite linear;
 }
@@ -74,4 +75,26 @@
   div.try_examples_button_container {
     display: none;
   }
+}
+
+/* Some improvements for hidden buttons */
+
+/* smooth transitions */
+.try_examples_button {
+  opacity: 1;
+  transition: opacity 0.3s ease-in-out;
+}
+
+.try_examples_button.hidden {
+  display: none;
+  opacity: 0;
+}
+
+.try_examples_button.fade-in {
+  transition: opacity 0.3s ease-in-out;
+}
+
+/* to ensure hidden class takes precedence */
+.hidden {
+  display: none !important;
 }
--- a/jupyterlite_sphinx/jupyterlite_sphinx.js
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.js
@@ -185,7 +185,7 @@ window.loadTryExamplesConfig = async (configFilePath) => {
     Patterns = data.ignore_patterns;
     for (let pattern of Patterns) {
       let regex = new RegExp(pattern);
-      if (!regex.test(currentPageUrl)) {
+      if (regex.test(currentPageUrl)) {
         var buttons = document.getElementsByClassName(
           "try_examples_button hidden",
         );
--- a/jupyterlite_sphinx/jupyterlite_sphinx.js
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.js
@@ -180,15 +180,18 @@ window.loadTryExamplesConfig = async (configFilePath) => {
       tryExamplesGlobalMinHeight = parseInt(data.global_min_height);
     }
 
-    // Disable interactive examples if file matches one of the ignore patterns
-    // by hiding try_examples_buttons.
+    // Selectively enable interactive examples if file matches one of the ignore patterns
+    // by un-hiding try_examples_buttons.
     Patterns = data.ignore_patterns;
     for (let pattern of Patterns) {
       let regex = new RegExp(pattern);
-      if (regex.test(currentPageUrl)) {
-        var buttons = document.getElementsByClassName("try_examples_button");
+      if (!regex.test(currentPageUrl)) {
+        var buttons = document.getElementsByClassName(
+          "try_examples_button hidden",
+        );
         for (var i = 0; i < buttons.length; i++) {
-          buttons[i].classList.add("hidden");
+          console.log(buttons[i]);
+          buttons[i].classList.remove("hidden");
         }
         break;
       }
--- a/jupyterlite_sphinx/jupyterlite_sphinx.js
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.js
@@ -205,7 +205,7 @@ window.loadTryExamplesConfig = async (configFilePath) => {
 window.toggleTryExamplesButtons = () => {
   /* Toggle visibility of TryExamples buttons. For use in console for debug
    * purposes. */
-  var buttons = document.getElementsByClassName("try_examples_button");
+  var buttons = document.getElementsByClassName("try_examples_button hidden");
 
   for (var i = 0; i < buttons.length; i++) {
     buttons[i].classList.toggle("hidden");
--- a/jupyterlite_sphinx/jupyterlite_sphinx.js
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.js
@@ -163,8 +163,18 @@ window.loadTryExamplesConfig = async (configFilePath) => {
     const response = await fetch(configFileUrl);
     if (!response.ok) {
       if (response.status === 404) {
-        // Try examples ignore file is not present.
+        // Try examples ignore file is not present. Enable all interactive examples
+        // in that case.
         console.log("Optional try_examples config file not found.");
+
+        var buttons = document.getElementsByClassName(
+          "try_examples_button hidden",
+        );
+        for (var i = 0; i < buttons.length; i++) {
+          buttons[i].classList.remove("hidden");
+        }
+        tryExamplesConfigLoaded = true;
+
         return;
       }
       throw new Error(`Error fetching ${configFilePath}`);
@@ -195,11 +205,11 @@ window.loadTryExamplesConfig = async (configFilePath) => {
         }
         break;
       }
+      tryExamplesConfigLoaded = true;
     }
   } catch (error) {
     console.error(error);
   }
-  tryExamplesConfigLoaded = true;
 };
 
 window.toggleTryExamplesButtons = () => {
--- a/jupyterlite_sphinx/jupyterlite_sphinx.js
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.js
@@ -165,14 +165,26 @@ window.loadTryExamplesConfig = async (configFilePath) => {
       if (response.status === 404) {
         // Try examples ignore file is not present. Enable all interactive examples
         // in that case.
-        console.log("Optional try_examples config file not found.");
+        console.log(
+          "Optional try_examples config file not found. Enabling all interactive examples.",
+        );
 
-        var buttons = document.getElementsByClassName(
+        // Grab all hidden try examples buttons and fade them in
+        const buttons = document.getElementsByClassName(
           "try_examples_button hidden",
         );
-        for (var i = 0; i < buttons.length; i++) {
-          buttons[i].classList.remove("hidden");
-        }
+
+        requestAnimationFrame(() => {
+          for (let button of buttons) {
+            button.style.opacity = "0";
+            button.classList.remove("hidden");
+            button.classList.add("fade-in");
+            requestAnimationFrame(() => {
+              button.style.opacity = "1";
+            });
+          }
+        });
+
         tryExamplesConfigLoaded = true;
 
         return;
@@ -182,6 +194,17 @@ window.loadTryExamplesConfig = async (configFilePath) => {
 
     const data = await response.json();
     if (!data) {
+      // If config file exists but is empty, treat it like a missing config
+      // and enable all interactive examples since there is no ignore pattern
+      console.log(
+        "Try examples config file is empty. Enabling all interactive examples.",
+      );
+      const buttons = document.getElementsByClassName(
+        "try_examples_button hidden",
+      );
+      for (let button of buttons) {
+        button.classList.remove("hidden");
+      }
       return;
     }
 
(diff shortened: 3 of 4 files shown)


agriyakhetarpal:
I have rebased this right now and opened a PR because I think it will be better for visibility, plus I would appreciate some help towards bringing this to completion. It's not fully working right now since the directives still seem to contain the `try_examples_button hidden` and therefore are not displayed, even when I try to remove the keyword from the `classList`. It has been a while since I worked on this, so I'm not sure what I was debugging back then.

.gitignore
```diff
@@ -5,3 +5,4 @@ docs/build
 build
 dist
 .jupyterlite.doit.db
+/.vscode/
```
Carreau, line 8: I think this is unrelated, and in general I believe (personally, not the position of the project), that you shoudl use [global gitignores](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c). There is no reason for every project on the internet to accomodate ever IDE on the planet, and once you setup your global .gitignore, then you will never have the problem of commiting .vscode folder ever.

jupyterlite_sphinx/jupyterlite_sphinx.js
```diff
@@ -180,15 +180,18 @@ window.loadTryExamplesConfig = async (configFilePath) => {
...
     }
 
-    // Disable interactive examples if file matches one of the ignore patterns
-    // by hiding try_examples_buttons.
+    // Selectively enable interactive examples if file matches one of the ignore patterns
+    // by un-hiding try_examples_buttons.
     Patterns = data.ignore_patterns;
     for (let pattern of Patterns) {
       let regex = new RegExp(pattern);
       if (regex.test(currentPageUrl)) {
```
Carreau, line 188: Should the condition be inverted ?

if match add hidden -> if not match remove hidden ?

Carreau (commented):
I see why this is reasonable, but isn't that going to have the opposite effect on page where the button should be visible, where the page will load and then jump around as buttons are added ?

agriyakhetarpal:
> I see why this is reasonable, but isn't that going to have the opposite effect on page where the button should be visible, where the page will load and then jump around as buttons are added ?

Thanks for the review, @Carreau. Yes, this will be a slight issue when someone reloads a page and if they are in the middle of a page when doing so – but at the top of the page, it should load as normal, because from there the buttons will be appended to the layout rather than being inserted into a position where the text moves because of the insertion. I am not sure if there is a cleaner way to do this.

.gitignore
agriyakhetarpal: Thanks! TIL about a global `.gitignore` file and how to set it up. Very useful!

jupyterlite_sphinx/jupyterlite_sphinx.js
agriyakhetarpal: Oops, yes – reverted the change in 0e88cbf7ccf8b5eb7325e2ceeb616fdf8bdb9b54.

agriyakhetarpal:
I think this is partially working now – where it breaks is when a `try_examples.json` config file is not present, and therefore `data.ignore_patterns` will always be false. I'll try to move the logic for un-hiding the buttons into a separate function and then call it. This way, the currently hidden buttons can be un-hidden with or without the presence of the configuration file.

agriyakhetarpal:
I think this is working now, marking it as ready for review. One small thing to figure out remains: the PR preview at https://jupyterlite-sphinx--173.org.readthedocs.build/en/173/directives/try_examples.html is not un-hiding the button with the custom `blue-bottom` CSS present in `jupyterlite_sphinx.css`. The button is visible in the stable documentation: https://jupyterlite-sphinx.readthedocs.io/en/stable/directives/try_examples.html

So this works for all sorts of buttons now (tested locally), except for these non-standard buttons. Let me try to dig further, and meanwhile open this up for further comments.

Carreau:
I will try to re-review. 

I'm ok with this in principle. Still unsure if hiding or appending is the best, so I'd like someone else input on this anyway.

agriyakhetarpal:
> I think this is working now, marking it as ready for review. One small thing to figure out remains: the PR preview at [jupyterlite-sphinx--173.org.readthedocs.build/en/173/directives/try_examples.html](https://jupyterlite-sphinx--173.org.readthedocs.build/en/173/directives/try_examples.html) is not un-hiding the button with the custom `blue-bottom` CSS present in `jupyterlite_sphinx.css`. The button is visible in the stable documentation: [jupyterlite-sphinx.readthedocs.io/en/stable/directives/try_examples.html](https://jupyterlite-sphinx.readthedocs.io/en/stable/directives/try_examples.html)
> 
> So this works for all sorts of buttons now (tested locally), except for these non-standard buttons that have user-defined CSS. Let me try to dig further, and meanwhile open this up for further comments.

Update on this: With the refactoring I performed in my recent commits, I've fixed the problem I had noted above; however, using it on projects downstream, such as NumPy, fails, and I am not able to un-hide the buttons selectively based on the config file. So far, I still haven't been able to figure out why (it apparently has something to do with the CSS rather than with JS).

Even though this is marked as a draft, it's ready for review – @Carreau, could you please re-review it when you have time? I discussed this PR with @steppi, who agreed to review it soon after I suggested that I would revisit it, which I now have. Perhaps some trained pairs of eyes will be able to catch issues I haven't been able to :D

Also, a note on the jumping around of the contents when the page is loaded: I've tried to implement a solution (not pushed here, though, it should probably be a separate PR) to minimise the [CLS](https://web.dev/articles/cls) value, through pre-allocating space for the container where the button is going to show up. However, for pages where the buttons are ignored/hidden, this check happens at runtime through the JavaScript code – which is not ideal. To achieve a smooth experience and no shifts in the content in the truest sense, this is possible only if we read through the documentation source pages and exclude them based on ignore patterns at build time (i.e., which would be read within the `TryExamplesDirective` class on the Python side), rather than at runtime on the JavaScript side, so that we avoid adding the HTML for the buttons altogether in the first place for pages where we need to ignore adding examples. I managed to get started on it, but I'm yet to test it (and I feel that should be in a separate PR, too).

jupyterlite_sphinx/jupyterlite_sphinx.css
```diff
@@ -75,3 +76,25 @@
...
+}
+
+.try_examples_button.fade-in {
+  transition: opacity 0.3s ease-in-out;
+}
+
+/* to ensure hidden class takes precedence */
+.hidden {
+  display: none !important;
+}
```
gabalafou, line 100: I'm not entirely sure that it's worth creating the fade-in effect when there is also a content layout shift.  

But if we still want to keep it, I suggest a few changes:

```suggestion

/* Buttons are hidden by default and revealed via JavaScript in another file */

/* The button is revealed in two stages. The first stage removes the hidden class.
  This gives the button dimensions and causes the page content below to "jump" 
  (get pushed down suddenly). The second stage removes the transparent class,
  which reveals the button with a slight fade-in effect. */
.try_examples_button.transparent {
  opacity: 0;
}

.try_examples_button {
  opacity: 1;
  transition: opacity 0.3s ease-in-out;
}
```

jupyterlite_sphinx/jupyterlite_sphinx.js
```diff
@@ -163,15 +163,48 @@ window.loadTryExamplesConfig = async (configFilePath) => {
...
+        requestAnimationFrame(() => {
+          for (let button of buttons) {
+            button.style.opacity = "0";
+            button.classList.remove("hidden");
+            button.classList.add("fade-in");
+            requestAnimationFrame(() => {
+              button.style.opacity = "1";
+            });
+          }
+        });
```
gabalafou, line 186: Let's wrap this in a function:

```js
const showButtons = () => {
  const buttons = document.querySelectorAll(
    ".try_examples_button.hidden",
  );
  for (let button of buttons) {
    button.classList.remove("hidden");
  }
  // If the hidden and transparent classes are removed
  // one right after the other, there is no animation of the opacity 
  // going from 0 to 1, hence requestAnimationFrame.
  requestAnimationFrame(() => {
    for (let button of buttons) {
      button.classList.remove("transparent");
    }
  });
}
```

jupyterlite_sphinx/jupyterlite_sphinx.js
```diff
@@ -163,15 +163,48 @@ window.loadTryExamplesConfig = async (configFilePath) => {
...
+      // and enable all interactive examples since there is no ignore pattern
+      console.log(
+        "Try examples config file is empty. Enabling all interactive examples.",
+      );
+      const buttons = document.getElementsByClassName(
+        "try_examples_button hidden",
+      );
+      for (let button of buttons) {
+        button.classList.remove("hidden");
+      }
```
gabalafou, line 207:
```suggestion
      showButtons();
```

jupyterlite_sphinx/jupyterlite_sphinx.js
```diff
@@ -180,29 +213,51 @@ window.loadTryExamplesConfig = async (configFilePath) => {
...
+    let shouldShowButtons = true;
+
     for (let pattern of Patterns) {
       let regex = new RegExp(pattern);
       if (regex.test(currentPageUrl)) {
-        var buttons = document.getElementsByClassName("try_examples_button");
-        for (var i = 0; i < buttons.length; i++) {
-          buttons[i].classList.add("hidden");
-        }
+        shouldShowButtons = false;
```
gabalafou, line 224:
```suggestion
        showButtons();
```

jupyterlite_sphinx/jupyterlite_sphinx.js
```diff
@@ -180,29 +213,51 @@ window.loadTryExamplesConfig = async (configFilePath) => {
...
+        for (let button of buttons) {
+          button.style.opacity = "0";
+          button.classList.remove("hidden");
+          button.classList.add("fade-in");
+          requestAnimationFrame(() => {
+            button.style.opacity = "1";
+          });
+        }
+      });
+    }
```
gabalafou, line 243:
```suggestion
```

jupyterlite_sphinx/jupyterlite_sphinx.js
```diff
@@ -180,29 +213,51 @@ window.loadTryExamplesConfig = async (configFilePath) => {
...
     }
 
-    // Disable interactive examples if file matches one of the ignore patterns
-    // by hiding try_examples_buttons.
-    Patterns = data.ignore_patterns;
+    // Selectively enable interactive examples if file matches one of the ignore patterns
+    // by un-hiding try_examples_buttons with a smooth transition
+    const Patterns = data.ignore_patterns || [];
+    let shouldShowButtons = true;
+
```
gabalafou, line 220:
```suggestion
```

jupyterlite_sphinx/jupyterlite_sphinx.js
```diff
@@ -180,29 +213,51 @@ window.loadTryExamplesConfig = async (configFilePath) => {
...
   } catch (error) {
-    console.error(error);
+    // On error, enable all buttons as a fallback to maintain current behavior
+    console.error("Error loading try_examples config:", error);
+    const buttons = document.getElementsByClassName(
+      "try_examples_button hidden",
+    );
+    for (let button of buttons) {
+      button.classList.remove("hidden");
+    }
```
gabalafou, line 252:
```suggestion
    showButtons();
```

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -553,15 +553,16 @@ def run(self):
         )
 
         # Button with the onclick event to swap embedded notebook back to examples.
+        # This includes a 'hidden' class by default to hide until the page is loaded.
         go_back_button_html = (
-            '<button class="try_examples_button" '
+            '<button class="try_examples_button hidden" '
```
gabalafou, line 558: From what I can tell, these buttons are already within a `.hidden` container.
```suggestion
            '<button class="try_examples_button" '
```

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -553,15 +553,16 @@ def run(self):
...
             f"onclick=\"window.tryExamplesHideIframe('{examples_div_id}',"
             f"'{iframe_parent_div_id}')\">"
             "Go Back</button>"
         )
 
         full_screen_button_html = (
-            '<button class="try_examples_button" '
+            '<button class="try_examples_button hidden" '
```
gabalafou, line 565:
```suggestion
            '<button class="try_examples_button" '
```

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -570,7 +571,7 @@ def run(self):
         # Button with the onclick event to swap examples with embedded notebook.
         try_it_button_html = (
             '<div class="try_examples_button_container">'
-            '<button class="try_examples_button" '
+            '<button class="try_examples_button hidden" '
```
gabalafou, line 574:
```suggestion
            '<button class="try_examples_button hidden transparent" '
```

(see other suggestions above)

jupyterlite_sphinx/jupyterlite_sphinx.js
```diff
@@ -180,29 +213,51 @@ window.loadTryExamplesConfig = async (configFilePath) => {
...
   }
   tryExamplesConfigLoaded = true;
 };
 
 window.toggleTryExamplesButtons = () => {
   /* Toggle visibility of TryExamples buttons. For use in console for debug
    * purposes. */
-  var buttons = document.getElementsByClassName("try_examples_button");
+  var buttons = document.getElementsByClassName("try_examples_button hidden");
```
gabalafou, line 260: This is a toggle function, so you need to be able to find toggle any button whether or not it is already hidden.

```suggestion
  var buttons = document.getElementsByClassName("try_examples_button");
```

jupyterlite_sphinx/jupyterlite_sphinx.js
```diff
@@ -180,29 +213,51 @@ window.loadTryExamplesConfig = async (configFilePath) => {
...
 
   for (var i = 0; i < buttons.length; i++) {
     buttons[i].classList.toggle("hidden");
```
gabalafou, line 263:
```suggestion
    buttons[i].classList.toggle("hidden");
    requestAnimationFrame(() => {
      buttons[i].classList.toggle("transparent");
    });
```

gabalafou (changes requested):
Nice work to improve the user experience.

I agree that it's a very poor user experience to see a page with buttons and then see those buttons disappear inexplicably.

From that perspective, this pull request looks like it's doing the right thing.

However, if the paradigm is that most websites exclude only a small portion of pages, then this pull request penalizes the main case (include the page, show the buttons) in order to protect the exceptional case (ignore the page, hide the buttons).

I'm not sure there's really any good way to do this other than the one you mentioned—page exclusion patterns should be handled at build-time, not at run-time.

But we need to know why it was architected that way to begin with. There may be some really important use case for being able to dynamically add or remove exclusion patterns  without having to rebuild the docs.

I'm requesting changes because the CSS should not have an `!important` flag, the toggle function needs fixing I think, and the JavaScript should have less copy-pasted code by creating `showButtons()` function.

PR #180, opened 2024-06-26
Strip tagged cells from `.ipynb` notebooks passed to the `NotebookLite` and `JupyterLite` directives

agriyakhetarpal:
## Description

This PR updates the `_LiteDirective` class to be able to use `nbformat`, which is already specified as a dependency in `pyproject.toml`, to be able to read IPyNB files and remove cells which include "strip" set to "true" in their JSON metadata. This way, it is possible to provide cleaner notebooks for cases similar to https://github.com/scipy/scipy/pull/20303 or with https://github.com/PyWavelets/pywt/pull/741, where notebooks passed to the `.. notebooklite::` or the `.. jupyterlite::` directives can now be preprocessed to remove certain cells before they are copied to the built documentation.

For example, a notebook cell with the Markdown syntax

````markdown
```{note}
This text should not exist in the notebook in the built documentation
```
````

can receive metadata like this

```json
{
  "tags": [
    "strip": "true"
  ]
}
```

and this cell will be removed by `nbformat` if `strip_tagged_cells` is set to `True` in `conf.py`, therefore disabling the admonition in the deployed documentation.

## Changes made

1. Added a configuration value for the Sphinx extension, `strip_tagged_cells` which is set to `False` by default. If set to `True` in `conf.py`, this enables preprocessing behaviour
2. The preprocessing behaviour is in `jupyterlite_sphinx/jupyterlite_sphinx.py`, where `nbformat` is used to read the notebook file and operate on the cells of the notebook.
3. This feature has been documented in `docs/configuration.md`.

## Additional context

- The `TryExamples` directive is not compatible with this feature. This is because we need an IPyNB-based notebook file to add metadata for a cell, while `TryExamples` dynamically generates notebooks at runtime from doctests.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -64,12 +64,13 @@ This behaviour can be enabled by setting the following config:
 strip_tagged_cells = True
 ```
 
-and then tag the cells you want to strip with the tag `strip` in the JSON metadata of the cell, like this:
+and then tag the cells you want to strip with the tag `jupyterlite_sphinx_strip` in the JSON metadata
+of the cell, like this:
 
 ```json
 {
   "tags": [
-    "strip": "true"
+    "jupyterlite_sphinx_strip": "true"
   ]
 }
 ```
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -55,9 +55,10 @@ jupyterlite_config = "jupyterlite_config.json"
 
 ## Strip particular tagged cells from IPython Notebooks
 
-When using the `NotebookLite` or the `JupyterLite` directives with a notebook passed, you can add strip particular tagged cells from the notebook before rendering it in the JupyterLite console.
+When using the `NotebookLite`, `JupyterLite`, or `Voici` directives with a notebook passed, you can
+strip particular tagged cells from the notebook before rendering it in the JupyterLite console.
 
-You can enable this behaviour by setting the following config:
+This behaviour can be enabled by setting the following config:
 
 ```python
 strip_tagged_cells = True
@@ -73,9 +74,12 @@ and then tag the cells you want to strip with the tag `strip` in the JSON metada
 }
 ```
 
-This is useful when you want to remove some cells from the rendered notebook in the JupyterLite console, for example, cells that are used for adding reST-based directives or other Sphinx-specific content.
+This is useful when you want to remove some cells from the rendered notebook in the JupyterLite
+console, for example, cells that are used for adding reST-based directives or other
+Sphinx-specific content.
 
-For example, you can use this feature to remove the `toctree` directive from the rendered notebook in the JupyterLite console:
+For example, you can use this feature to remove the `toctree` directive from the rendered notebook
+in the JupyterLite console:
 
 ```json
 {
@@ -107,9 +111,12 @@ For example, you can use this feature to remove the `toctree` directive from the
 }
 ```
 
-where the cell with the `toctree` directive will be removed from the rendered notebook in the JupyterLite console.
+where the cell with the `toctree` directive will be removed from the rendered notebook in
+the JupyterLite console.
 
-Note that this feature is only available for the `NotebookLite` and `JupyterLite` directives and works with the `ipynb` files passed to them, and therefore does not work with the `TryExamples` directive.
+Note that this feature is only available for the `NotebookLite`, `JupyterLite`, and the
+`Voici` directives and works with the `ipynb` files passed to them, and therefore does
+not work with the `TryExamples` directive.
 
 ## Disable the `.ipynb` docs source binding
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -53,6 +53,64 @@ to your JupyterLite deployment.
 jupyterlite_config = "jupyterlite_config.json"
 ```
 
+## Strip particular tagged cells from IPython Notebooks
+
+When using the `NotebookLite` or the `JupyterLite` directives with a notebook passed, you can add strip particular tagged cells from the notebook before rendering it in the JupyterLite console.
+
+You can enable this behaviour by setting the following config:
+
+```python
+strip_tagged_cells = True
+```
+
+and then tag the cells you want to strip with the tag `strip` in the JSON metadata of the cell, like this:
+
+```json
+{
+  "tags": [
+    "strip": "true"
+  ]
+}
+```
+
+This is useful when you want to remove some cells from the rendered notebook in the JupyterLite console, for example, cells that are used for adding reST-based directives or other Sphinx-specific content.
+
+For example, you can use this feature to remove the `toctree` directive from the rendered notebook in the JupyterLite console:
+
+```ipynb
+{
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "tags": [
+          "strip": "true"
+        ]
+      },
+      "source": [
+        "# Table of Contents\n",
+        "\n",
+        "```{toctree}\n",
+        ":maxdepth: 2\n",
+        "\n",
+        "directives/jupyterlite\n",
+        "directives/notebooklite\n",
+        "directives/replite\n",
+        "directives/voici\n",
+        "directives/try_examples\n",
+        "full\n",
+        "changelog\n",
+        "```"
+      ]
+    }
+  ]
+}
+```
+
+where the cell with the `toctree` directive will be removed from the rendered notebook in the JupyterLite console.
+
+Note that this feature is only available for the `NotebookLite` and `JupyterLite` directives and works with the `ipynb` files passed to them, and therefore does not work with the `TryExamples` directive.
+
 ## Disable the `.ipynb` docs source binding
 
 By default, jupyterlite-sphinx binds the `.ipynb` source suffix so that it renders Notebooks included in the doctree with JupyterLite.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -77,7 +77,7 @@ This is useful when you want to remove some cells from the rendered notebook in 
 
 For example, you can use this feature to remove the `toctree` directive from the rendered notebook in the JupyterLite console:
 
-```ipynb
+```json
 {
   "cells": [
     {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -55,7 +55,7 @@ jupyterlite_config = "jupyterlite_config.json"
 
 ## Strip particular tagged cells from IPython Notebooks
 
-When using the `NotebookLite`, `JupyterLite`, or `Voici` directives with a notebook passed, you can
+When using the `NotebookLite`, `JupyterLite`, or `Voici` directives with a notebook passed to them, you can
 strip particular tagged cells from the notebook before rendering it in the JupyterLite console.
 
 This behaviour can be enabled by setting the following config:
@@ -64,7 +64,7 @@ This behaviour can be enabled by setting the following config:
 strip_tagged_cells = True
 ```
 
-and then tag the cells you want to strip with the tag `jupyterlite_sphinx_strip` in the JSON metadata
+and then by tagging the cells you want to strip with the tag `jupyterlite_sphinx_strip` in the JSON metadata
 of the cell, like this:
 
 ```json
@@ -89,7 +89,7 @@ in the JupyterLite console:
       "cell_type": "markdown",
       "metadata": {
         "tags": [
-          "strip": "true"
+          "jupyterlite_sphinx_strip": "true"
         ]
       },
       "source": [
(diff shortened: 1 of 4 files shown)


agriyakhetarpal:
In commit 23d2cae06579740fbfda9e17396a91ddfef4857f, I added an example by setting `strip_tagged_cells = True` and modified `my_notebook.ipynb` accordingly – the notebook in the hosted documentation contains two cells, while the one in the source has three – which means that one of the cells (the tagged one) was removed.

This can be noticed on the pages in the documentation preview: https://jupyterlite-sphinx--180.org.readthedocs.build/en/180/directives/jupyterlite.html and https://jupyterlite-sphinx--180.org.readthedocs.build/en/180/directives/notebooklite.html.

agriyakhetarpal:
@steppi, I figured out the problem with the repeated copying for the notebook files, and the good thing is that the repeated copying exists only when the notebook is embedded multiple times in the same document (I'm appalled it took me slightly longer than expected to figure it out, haha). This behaviour is apparently different from that of the `TryExamples` directive in that individual `.ipynb` notebooks will likely, in real-life documentation scenarios, be referenced just once in a singleton directive somewhere, and therefore we should not need UUIDs at all, far apart the case of the `TryExamples` directive. Further, it is unrealistic that two notebooks will be placed somewhere in the source for any project's documentation with the exact same filenames. It was only when I tried the experiment for Melissa below that I understood from the logs that `my_notebook.ipynb` is referenced multiple times in the `jupyterlite-sphinx` docs, and therefore is copied multiple times (which may be fixed with the `env.temp_data` solution you suggested, but I would suggest we leave that optimisation for another PR), while the rest of the notebooks are copied just once, as intended.

@melissawm, I tested this with the seventeen or so notebooks from https://github.com/scipy/scipy/pull/21042 like you suggested (I converted them via Jupytext to IPyNB first, using its CLI) and added them under separate JupyterLite directives. The time taken by `nbformat` to process the notebooks was not noticeable at all!

<hr>

P.S. Jupytext provides an API as well. Perhaps another worthwhile feature to add to `jupyterlite-sphinx` would be to allow `.. notebooklite::` and `..jupyterlite::` to accept these text-based Markdown notebooks? Jupytext can then convert the notebook from `.md` at runtime here itself in these changed lines (maybe even by default), and pass/copy it as `.ipynb` to JupyterLite – this can potentially resolve the problem of having extra targets/phonies in the docs Makefiles implemented in https://github.com/scipy/scipy/pull/20303 and also removes the addition of a custom Sphinx extension connecting to the `builder-inited` event in https://github.com/PyWavelets/pywt/pull/741. This should also help make the docs directory cleaner at the time of rebuilds, since Markdown-based notebooks will not be required to be paired with their IPyNB counterparts (which may have been git-ignored or excluded from inclusion in Sphinx). 

It is to be noted that the JupyterLite interface does have support for Jupytext in the screenshot below, but I am not sure how to make the Markdown notebook runnable – (I think) it is not supported right now: https://github.com/jupyterlite/jupyterlite/discussions/731, so this conversion from `.md` to `.ipynb` at build time could be a reasonable solution for notebook authors until JupyterLite itself can gain support.

<details>
<summary>"Jupytext" under the File tab under the standard JupyterLite interface</summary>

Jupytext, hovered over in the File tab in the JupyterLite interface, local Sphinx docs for the jupyterlite-sphinx project's JupyterLite directive that embeds notebook files

</details>

Carreau:
Should this option be a booolean or a list of tag to strip?

```
cell_strip_tag = ['solutions', 'strip', 'debug']
```

?

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -338,10 +340,45 @@ def run(self):
...
+                    print(f"Opened {notebook_name}")
+                    nb = nbformat.read(f, as_version=4)
+                    nb.cells = [
+                        cell
+                        for cell in nb.cells
+                        if "true" not in cell.metadata.get("strip", [])
+                    ]
+                    print(f"Writing stripped notebook to {temp_dir}")
+                    nbformat.write(nb, Path(temp_dir) / notebook_name)
+
```
Carreau, line 374: Can't you directly read from `notebook`, and write to `notebooks_dir` ? If possible I'd like to not touch the filesystem twice.

It may requires moving `os.makedirs(os.path.dirname(notebooks_dir), exist_ok=True)` earlier.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -338,10 +340,45 @@ def run(self):
 
             notebooks_dir = Path(self.env.app.srcdir) / CONTENT_DIR / notebook_name
 
+            # Create a temporary directory and copy the notebooks there. This is done
+            # internally so that we don't overwrite existing notebooks in the source
+            # directories of the docs (if we are modifying their contents). Therefore
+            # we copy from an intermediate directory to the final one where the notebooks
+            # are expected to be found.
+
+            import tempfile
```
Carreau, line 349: Likely move import to top of file.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -338,10 +340,45 @@ def run(self):
...
+
+            temp_dir = tempfile.mkdtemp()
+            # Copy notebooks in notebooks_dir to temp_dir
+            shutil.copy(notebook, temp_dir)
+
+            if self.env.config.strip_tagged_cells:
+                print(
```
Carreau, line 356: not for this PR, but I belive Directive have a `.info()`, `.warning(...)` methods maybe.

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: Removed `tempfile` entirely in 31466befd0e5f288d924ddefa1be18596d08abbc

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: Yes, it required moving `os.makedirs(os.path.dirname(notebooks_dir), exist_ok=True)` earlier. Addressed in 31466befd0e5f288d924ddefa1be18596d08abbc, now, I removed the temporary directory stuff altogether – this should be cleaner.

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: Ah, I was using these `print()` statements for a bit of debugging here and there, do you want me to remove them for this PR?

agriyakhetarpal:
> Should this option be a booolean or a list of tag to strip?
> 
> ```
> cell_strip_tag = ['solutions', 'strip', 'debug']
> ```
> 
> ?

I have put some thought about the use cases of having JSON metadata for notebooks, e.g., to include interactive widgets like Thebe, or cells where it is helpful to show inputs but hide outputs, etc., so, yes – having a list of tags could be a reasonable option. I imagine this option will be better coupled with `strip_tagged_cells` instead of replacing it, though – i.e., have a boolean `strip_tagged_cells` specify whether to strip these cells, and then `cell_strip_tag` lists which tags to look for when stripping. What do you think, @Carreau?

Carreau:
> What do you think, @Carreau?

I don't have much preference. I tend to think that if it's two options it might be more difficult to configure, and if the list is empty it can play the role of not stripping ?

I think I would be +0.1 for just having a single config option. But maybe others want to pitch in.

agriyakhetarpal:
@steppi, do you have any additional comments to what @Carreau has mentioned? It would be great if you could take a look at this soon!

steppi:
> @steppi, do you have any additional comments to what @Carreau has mentioned? It would be great if you could take a look at this soon!

Sure, I'll take a look now.

jupyterlite_sphinx/jupyterlite_sphinx.py
steppi: I'm in favor of removing the raw `print`s from this PR.

docs/configuration.md
```diff
@@ -53,6 +53,71 @@ to your JupyterLite deployment.
...
+```
+
+and then tag the cells you want to strip with the tag `strip` in the JSON metadata of the cell, like this:
+
+```json
+{
+  "tags": [
+    "strip": "true"
+  ]
+}
```
steppi, line 74: What do you think about changing the tag name to `jupyterlite_sphinx_strip`? Then if someone sees one of these notebooks in the wild, there will be more context to figure out what the tag means, and a google search would probably even pull up the `jupyterlite_sphinx` documentation.

steppi (commented):
This looks good @agriyakhetarpal. I'm in favor of keeping only a boolean config option and not making this more complex than needed. I also have a suggestion about the name of the tag.

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: Addressed in a50e099a754788e7133374a072cdcb7d03cdd7f5.

docs/configuration.md
agriyakhetarpal: I agree. Didn't think about how it would show up in a Google search, nice idea.

agriyakhetarpal: Addressed in 2af69e5c3655713bb99ce7579a56760a923ec4a6.

docs/configuration.md
```diff
@@ -53,6 +53,72 @@ to your JupyterLite deployment.
 jupyterlite_config = "jupyterlite_config.json"
 ```
 
+## Strip particular tagged cells from IPython Notebooks
+
+When using the `NotebookLite`, `JupyterLite`, or `Voici` directives with a notebook passed, you can
```
agriyakhetarpal, line 58:
```suggestion
When using the `NotebookLite`, `JupyterLite`, or `Voici` directives with a notebook passed to them, you can
```

docs/configuration.md
```diff
@@ -53,6 +53,72 @@ to your JupyterLite deployment.
...
+strip particular tagged cells from the notebook before rendering it in the JupyterLite console.
+
+This behaviour can be enabled by setting the following config:
+
+```python
+strip_tagged_cells = True
+```
+
+and then tag the cells you want to strip with the tag `jupyterlite_sphinx_strip` in the JSON metadata
```
agriyakhetarpal, line 67:
```suggestion
and then by tagging the cells you want to strip with the tag `jupyterlite_sphinx_strip` in the JSON metadata
```

docs/configuration.md
```diff
@@ -53,6 +53,72 @@ to your JupyterLite deployment.
...
+in the JupyterLite console:
+
+```json
+{
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "tags": [
+          "strip": "true"
```
agriyakhetarpal, line 92:
```suggestion
          "jupyterlite_sphinx_strip": "true"
```

docs/configuration.md
```diff
@@ -53,6 +53,72 @@ to your JupyterLite deployment.
...
+  ]
+}
+```
+
+where the cell with the `toctree` directive will be removed from the rendered notebook in
+the JupyterLite console.
+
+Note that this feature is only available for the `NotebookLite`, `JupyterLite`, and the
+`Voici` directives and works with the `ipynb` files passed to them, and therefore does
+not work with the `TryExamples` directive.
```
agriyakhetarpal, line 120:
```suggestion
Note that this feature is only available for the `NotebookLite`, `JupyterLite`, and the
`Voici` directives and works with the `.ipynb` files passed to them. It is not implemented
for the `TryExamples` directive.
```

docs/directives/my_notebook.ipynb
```diff
@@ -23,7 +23,9 @@
    "cell_type": "code",
    "execution_count": 2,
    "id": "84361143",
-   "metadata": {},
+   "metadata": {
+    "strip": "true"
+   },
```
agriyakhetarpal, line 28:
```suggestion
   "metadata": {},
```

docs/directives/my_notebook.ipynb
```diff
@@ -34,7 +36,8 @@
     }
    ],
    "source": [
-    "foo()"
+    "foo()\n",
+    "# This cell contains \"strip\": \"true\" in the metadata and therefore won't exist"
```
agriyakhetarpal, line 40:
```suggestion
    "foo()"
```

docs/directives/my_notebook.ipynb
```diff
@@ -43,7 +46,9 @@
    "id": "b9f8bed0",
    "metadata": {},
    "outputs": [],
-   "source": []
+   "source": [
+    "# This cell will exist because it hasn't been tagged"
+   ]
```
agriyakhetarpal, line 51:
```suggestion
   "source": []
```

agriyakhetarpal:
Reverted the changes I made to the notebook in b85346296fa7b4c1711138b6d660264c9c753e9e since the functionality is working as expected.

steppi (approved):
This looks good to me now. Thanks @agriyakhetarpal!

PR #187, opened 2024-07-16
Resolve default-encoding errors on Windows

AA-Turner:
See [``PYTHONWARNDEFAULTENCODING`` et al](https://docs.python.org/3/library/io.html#opt-in-encodingwarning). Without this change I ran into charmap errors. 

A

### DIFF
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -522,7 +522,9 @@ class TryExamplesDirective(SphinxDirective):
             ] = notebook_unique_name
             # Copy the Notebook for NotebookLite to find
             os.makedirs(notebooks_dir, exist_ok=True)
-            with open(notebooks_dir / Path(notebook_unique_name), "w", encoding="utf-8") as f:
+            with open(
+                notebooks_dir / Path(notebook_unique_name), "w", encoding="utf-8"
+            ) as f:
                 # nbf.write incorrectly formats multiline arrays in output.
                 json.dump(nb, f, indent=4, ensure_ascii=False)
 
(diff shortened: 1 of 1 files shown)


jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -522,7 +522,7 @@ def run(self):
             ] = notebook_unique_name
             # Copy the Notebook for NotebookLite to find
             os.makedirs(notebooks_dir, exist_ok=True)
-            with open(notebooks_dir / Path(notebook_unique_name), "w") as f:
+            with open(notebooks_dir / Path(notebook_unique_name), "w", encoding="utf-8") as f:
                 # nbf.write incorrectly formats multiline arrays in output.
                 json.dump(nb, f, indent=4, ensure_ascii=False)
```
AA-Turner, line 527: An alternative, given you're using Pathlib:
```suggestion
            # nbf.write incorrectly formats multiline arrays in output.
            nb_json = json.dump(nb, f, indent=4, ensure_ascii=False)
            (notebooks_dir / notebook_unique_name).write_text(nb_json, encoding="utf-8")
```

agriyakhetarpal: We have `nbformat` as a dependency now. Can we make it write the notebook here instead of having to rely on JSON directly?

agriyakhetarpal (approved):

jupyterlite_sphinx/jupyterlite_sphinx.py
AA-Turner: Yep sorry - I just made the suggestion via GH's editor, missed checking it properly.

A

steppi:
Thanks for the patch @AA-Turner. This project enforces the black codestyle; to fix the test failure you just have to run black on the file `jupyterlite_sphinx.py`.

AA-Turner:
Hi -- this was just a quick fix as I'm debugging a new failure in Sphinx 7.4/SciPy -- I don't have the project checked out locally so can't easily run black/etc. Please feel free to push to this branch / whatever else is appropriate.

A

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: Ah, thanks for the context, @steppi. I thought the problem might have been resolved by now, but that's something to look for in a different place.

steppi:
Thanks again for the fix @AA-Turner. I've made the formatting change so in this goes.

PR #189, opened 2024-07-18
Fix invalid schema for `jupyterlite_sphinx_strip` tag when `strip_tagged_cells` is `True`

agriyakhetarpal:
## Description

```json
"metadata": {
        "tags": [
          "jupyterlite_sphinx_strip": "true"
        ]
      }
```

is invalid for notebook-based metadata, while `"jupyterlite_sphinx_strip"` is valid. This was incorrectly mentioned in the documentation and was not caught because we were trying to parse the cell metadata itself. We now check for the existence of a tag in the metadata rather than trying to find it as a key-value pair in the metadata for a cell. This is a fix that is a candidate for a v0.16.1 release.

<hr>

cc: @steppi

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -68,9 +68,9 @@ and then by tagging the cells you want to strip with the tag `jupyterlite_sphinx
 of the cell, like this:
 
 ```json
-{
+"metadata": {
   "tags": [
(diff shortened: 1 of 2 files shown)


steppi (approved):
Looks good @agriyakhetarpal. Thanks for catching this early.

PR #190, opened 2024-07-18
Restore SameFileError check when copying notebooks

melissawm:
I think this is all we need to close https://github.com/scipy/scipy/issues/21212 -let me know if there's any follow-up. Cheers!


steppi (approved):
Thanks @melissawm. Looks good.

PR #193, opened 2024-07-24
Allow for JupyterLite 0.4.0

jtpio:
Prepare for the upcoming JupyterLite 0.4.0 release, by relaxing the upper bound on `jupyterlite-core`.

Normally `jupyterlite-sphinx` should now be affected by this new version, so we should allow for the latest `jupyterlite-core`.

Also allow for the next version of `jupyterlite-xeus` in the docs environment.


martinRenou (approved):
Thanks!

PR #199, opened 2024-08-05
Fix compatibility with Sphinx 8

### LINKED ISSUE #197: Compatibilty with Sphinx 8?
The CI doc builds for SciPy recently started failing: https://github.com/scipy/scipy/issues/21323

The error message suggests a problem with `jupyterlite_sphinx`.  It looks like an exception is being generated when used with the recently released Sphinx 8.  Is this a known issue?

agriyakhetarpal:
## Description

The function `_process_docstring_examples` used `app.env.doc2path`, which now needs paths instead of strings, since Sphinx 9 will drop support for representing paths as strings, which was raising warnings for the new release of Sphinx, i.e., version 8. This uses `pathlib.Path()` to process the path. A few other type hints have been added to the function's arguments.

Closes #197

## Additional context

xref docs build failures in SciPy downstream: https://github.com/scipy/scipy/issues/21323, https://github.com/scipy/scipy/pull/21324

### DIFF
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -4,7 +4,7 @@ from uuid import uuid4
 import shutil
 import glob
 import re
-from typing import Dict, Any
+from typing import Dict, Any, List
 
 from pathlib import Path
 
@@ -608,9 +608,9 @@ class TryExamplesDirective(SphinxDirective):
         return [content_container_node, notebook_container, script_node]
 
 
-def _process_docstring_examples(app, docname, source):
-    source_path = app.env.doc2path(docname)
-    if source_path.endswith(".py"):
+def _process_docstring_examples(app: Sphinx, docname: str, source: List[str]) -> None:
+    source_path: os.PathLike = Path(app.env.doc2path(docname))
+    if source_path.suffix == ".py":
         source[0] = insert_try_examples_directive(source[0])
 
 
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -609,8 +609,8 @@ class TryExamplesDirective(SphinxDirective):
 
 
 def _process_docstring_examples(app: Sphinx, docname: str, source: List[str]) -> None:
-    source_path: os.PathLike = Path(app.env.doc2path(docname))
-    if source_path.suffix == ".py":
+    source_path: str = str(app.env.doc2path(docname))
+    if source_path.endswith(".py"):
         source[0] = insert_try_examples_directive(source[0])
 
 
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -609,8 +609,8 @@ class TryExamplesDirective(SphinxDirective):
 
 
 def _process_docstring_examples(app: Sphinx, docname: str, source: List[str]) -> None:
-    source_path: str = str(app.env.doc2path(docname))
-    if source_path.endswith(".py"):
+    source_path: str = app.env.doc2path(docname)
+    if source_path.suffix == ".py":
         source[0] = insert_try_examples_directive(source[0])
 
 


agriyakhetarpal:
Thanks for flagging the issue, @WarrenWeckesser! This is ready for review.

WarrenWeckesser:
Thanks for the quick fix, @agriyakhetarpal.  When I locally replaced `jupyterlite_sphinx` 0.16.3 with this PR and built the SciPy docs, I no longer got the error from the `jupiterlite_sphinx` code.  I still got an error, but now it is coming from `myst_nb`.  So I suspect this fix exposes another package that needs to be updated.

steppi:
Looking into it a little. I think this is actually a Sphinx bug which has been fixed at least in main. It would be weird if Sphinx wouldn't accept it's own output from `doc2path`, and if you look at the latest docs, it now returns a `_StrPath`, which is a pathlib path which has string methods.

https://www.sphinx-doc.org/en/master/_modules/sphinx/environment.html#BuildEnvironment.doc2path

WarrenWeckesser:
FYI: I created an issue at `myst_nb`: https://github.com/executablebooks/MyST-NB/issues/619.  As I explain there, I can get the SciPy docs to build if I apply the same fix as used here to one line in the `myst_nb` code.

agriyakhetarpal:
Thanks for trying this out with MyST-NB, @WarrenWeckesser! I think we should be good to go with this PR, then, @steppi – since the warning is no longer emitted here. The MyST-NB one could be suppressed if the fix there takes longer to land.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -608,9 +608,9 @@ def run(self):
         return [content_container_node, notebook_container, script_node]
 
 
-def _process_docstring_examples(app, docname, source):
-    source_path = app.env.doc2path(docname)
-    if source_path.endswith(".py"):
+def _process_docstring_examples(app: Sphinx, docname: str, source: List[str]) -> None:
+    source_path: os.PathLike = Path(app.env.doc2path(docname))
```
steppi, line 612: You shouldn't have to convert the output of `app.env.doc2path` into a `Path`. `app.env.doc2path` should return a `_StrPath`, which inherits from `Path` but adds string methods to it.

agriyakhetarpal: That makes sense. I noticed that they changed it to convert to `pathlib.Path()` in the release candidate for v8 but rolled it back in the final v8 release, trimming it down to a deprecation warning for further removal in v9. We can handle that later.

I'll convert the `_StrPath` object to a string instead based on how other projects seem to have fixed it:

```suggestion
    source_path: str = str(app.env.doc2path(docname))
```

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -608,9 +608,9 @@ def run(self):
...
+    if source_path.suffix == ".py":
```
agriyakhetarpal, line 613:
```suggestion
    if source_path.endswith(".py"):
```

jupyterlite_sphinx/jupyterlite_sphinx.py
steppi: I think it's better just to keep it a `Path` and use a `Path` method. This is my read on what the `Sphinx` developers would prefer. It seems like they're trying to be opinionated and say people should be using `pathlib` to handle paths.

agriyakhetarpal: I see – do you mean I should use `app.env.doc2path(docname)` as-is, and use `Path.suffix` in the following line?

I imagine downstream extension authors like us would agree that using `pathlib` would be broadly acceptable, though – I find it very elegant!

steppi: Yes, use the output of `app.env.doc2path` directly and use `Path.suffix`.  In `sphinx 9`, the output of `app.env.doc2path` will be a `Path`

agriyakhetarpal: Addressed in 13a3631b8e3f52e232a910eb284e6597f94894c3.

steppi (approved):
lgtm

PR #201, opened 2024-08-07
Restore backwards compatibility with Sphinx <8

agriyakhetarpal:
## Description

See https://github.com/numpy/numpy/pull/26745#issuecomment-2273605967 where I encountered this. #199 ensured compatibility with Sphinx 8 after https://github.com/scipy/scipy/issues/21323 was opened (temporarily pinned in https://github.com/scipy/scipy/pull/21324), but it seems to have broken compatibility with older versions of Sphinx (6 and 7, for example). @steppi, I have restored the original fix from #199, since I can't find older versions of the Sphinx documentation to read through – it seems that they publish just the version from the `master` branch. These changes should work with both Sphinx 8 and earlier versions.


Carreau:
working on a release, I just have issues with the pre-release workflow that does the wrong changelog. Investigating.

PR #204, opened 2024-08-20
chore: remove `import *` statements from `bqplot.ipynb`

agriyakhetarpal:
## Description

A tiny PR that closes #203; removes the `import *` statements from the `bqplot` example notebook. They are discouraged since they mix-up the namespace with too many imports. We could consider adding https://github.com/asmeurer/removestar as a pre-commit hook, but I don't think it's needed too much because this was the only failing example we had.


agriyakhetarpal:
This has been fixed in `main`, no longer needed

Issue #191, opened 2024-07-19
Support Jupytext (MyST-Markdown flavoured) notebooks in `NotebookLite`, `JupyterLite`, and `Voici` directives


agriyakhetarpal:
### Problem

`jupyterlite-sphinx` does not yet support Jupytext (Markdown) notebooks in its directives. Markdown notebooks are a wonderful solution to existing notebook tooling designed for IPyNB notebooks, which are hard to conquer for traditional version control systems like Git. One of the advantages of notebooks as text-based formats is that they pair up nicely with Markdown-based documentation tooling along with Sphinx and are supported for execution and for other standards in the ecosystem as well.

### Proposed Solution

This feature request is copied over from the proposal added in https://github.com/jupyterlite/jupyterlite-sphinx/pull/180#issuecomment-2192628783 as a corollary to the functionality provided and now available since [`v0.16.0` (release)](https://github.com/jupyterlite/jupyterlite-sphinx/releases/tag/v0.16.0):

The Jupyter extension, Jupytext, provides an API in addition to its CLI. Another worthwhile feature to add to `jupyterlite-sphinx` would be to allow the `.. notebooklite::`, `..jupyterlite::`, and the `..voici::` directives to accept text-based Markdown notebooks – `jupytext` as a dependency can then be used with its API to convert the notebook from `.md` at `jupyterlite-sphinx`'s runtime (maybe even by default, and not opt-in), and then pass/copy it as `.ipynb` to JupyterLite.

It is to be noted that the JupyterLite interface does have support for Jupytext, but I am not sure if the Markdown notebook is currently runnable – it is not supported right now, based on this discussion: https://github.com/jupyterlite/jupyterlite/discussions/731, so this conversion from `.md` to `.ipynb` at build time could be a reasonable _temporary_ solution for notebook authors until JupyterLite itself can gain support. We can instruct users in the documentation and let them be aware through the means of an admonition that their notebooks are available in the JupyterLite interface as IPyNB, even if they pass them in their directives as Markdown. Some use cases are available in the "Additional context" section below.

### Additional context

- `jupyterlite-sphinx` is increasingly used in Scientific Python documentation, as are Markdown notebooks. This feature can potentially resolve the problem of having extra targets/phonies in the docs Makefiles implemented in https://github.com/scipy/scipy/pull/20303, providing advantageous cleanup; and
[...]

PR #221, opened 2024-12-03
Support the usage of Markdown-based notebooks with "Lite" directives

agriyakhetarpal:
## Description

Please refer to the proposal in #191 for additional context and the rationale behind this change. This PR adds preprocessing functionalities to the base "Lite" directive (in the `_LiteIframe` class), which will allow a user to pass in a plain Markdown/MyST-Markdown notebook file to a directive, which will then get converted on the fly when JupyterLite is triggered. The conversion is performed via `jupytext`, which has been included as a dependency.

There is also the option of leaving this as an experimental feature behind an experimental flag and/or an optional dependency on `jupytext`, as suggested in https://github.com/jupyterlite/jupyterlite-sphinx/issues/191#issuecomment-2244589977. However, I haven't implemented that change and would like to wait for further direction on this before I do so.

## Changes made

- Preprocessing added to the `run()` method of the `_LiteIframe` class to convert `.md` files to `.ipynb` files and pass them on to JupyterLite
- Refactored the notebook conversion and cell stripping logic to two separate hidden functions that we subsequently call in `_LiteIframe.run()`
- Handled potential notebook collisions (in case notebooks of the same name but with different formats are present anywhere in the docs) – here, we handle only those notebooks that are referenced in a directive, and we raise a `ValueError` for any conflicts, recommending that the user rename their notebooks to have unique names.
- `jupytext` added as a dependency for facilitating the conversion
- Documentation for all directives updated to showcase this functionality
- A sample MyST Markdown notebook added

## Additional context

This is supposed to make interacting with Markdown-based notebook workflows easier – a few notable examples are scipy/scipy#20303 and PyWavelets/pywt#741, which have their own ways of dealing with this intermediate step that might make sense to include in `jupyterlite-sphinx` itself through this PR, such as https://github.com/scipy/scipy/blob/82ee02556872b9f91576b835a8f40f1d9eb1fd8d/doc/convert_notebooks.py and https://github.com/agriyakhetarpal/pywt/blob/fd3e01c3003a18cca770fc6f33752f8c07084e21/doc/source/conf.py#L33-L49 respectively.

cc: @melissawm @steppi @gabalafou


agriyakhetarpal:
I am facing some unrelated troubles when testing it locally with the SciPy documentation from SciPy's `main` branch right now, and I intend to test this a bit more thoroughly with multiple projects, also to ensure that things like https://github.com/scipy/scipy/pull/21433 do not break with this change.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -336,37 +337,51 @@ def run(self):
...
-            if notebook_is_stripped:
-                # Note: the directives meant to be stripped must be inside their own
-                # cell so that the cell itself gets removed from the notebook. This
-                # is so that we don't end up removing useful data or directives that
-                # are not meant to be removed.
+            # For MyST Markdown notebooks, we create a unique target filename to avoid
+            # collisions with other IPyNB files that may have the same name.
+            if notebook_path.suffix.lower() == '.md':
+                target_name = f"{notebook_path.stem}_{uuid4().hex[:8]}.ipynb"
+                target_path = notebooks_dir / target_name
```
agriyakhetarpal, line 348: Here, this is an assumption I make, which, while would be true for the `jupyterlite-sphinx` documentation because I've added a file `my_notebook.md` of the same name to accompany the IPyNB file – is a bit unlikely to be true for downstream projects:

- the notebooks would either be all IPyNB files, or all Markdown-based, or some other format entirely
- even if the notebooks are undergoing a migration from one format to another by a contributor such (as is the case for SciPy), it's unlikely that two notebooks with the same name but different formats, `foo.ipynb` and `foo.md`, would exist under the same folder (or different folders for the same matter).

If I am right in my assessment, we can drop the use of the UUID here and go ahead with these assumptions. Could someone help validate?

agriyakhetarpal:
> I am facing some unrelated troubles when testing it locally with the SciPy documentation from SciPy's `main` branch right now, and I intend to test this a bit more thoroughly with multiple projects, also to ensure that things like [scipy/scipy#21433](https://github.com/scipy/scipy/pull/21433) do not break with this change.

Update: I've tested it multiple times by now and this is working perfectly well with the SciPy docs 🎉 though the use of a UUID to append to the filename would break the caching (obviously) of the notebook builds and also adds additional notebooks to the built docs. So, I've chosen to adapt the caching-related check from https://github.com/scipy/scipy/pull/21433 in e3b6882b9dc25a897308a2b77bf8b3987a2fd7ed, and I can confirm that there are no regressions as those that were noticed in scipy/scipy#21394. This does mean that I've had to revert the commit via d1c7f7dd7aec734f55ddd909ee182c342e0ac86d where I added a sample `my_notebook.md` file to accompany our existing `my_notebook.ipynb` file, because of the naming conflict that will be caused after conversion (which is something I'm looking for opinions on – please see https://github.com/jupyterlite/jupyterlite-sphinx/pull/221#discussion_r1866936923 for context!)

When this PR becomes part of a release, we can remove the `convert_notebooks.py` script and the `convert-notebooks` Make target in SciPy, essentially eliminating the changes introduced through scipy/scipy#21433, and instead of having to write:

````rst
.. jupyterlite:: ../../_contents/hypothesis_chi2_contingency.ipynb
   :new_tab: True
````

we can just write:

```rst
.. jupyterlite:: hypothesis_chi2_contingency.md
   :new_tab: True
```

which would be much cleaner – and I can submit a follow-up PR to SciPy for this, or step back if you wish to do so, @melissawm. :) This PR should now be ready for review now!

agriyakhetarpal:
Note; we can also allow both `notebook_X.ipynb` and `notebook_X.md` to coexist in the same folder, and just add a suffix for the Markdown-based case, so that using `.. notebooklite:: notebook_X.ipynb` opens the original notebook, but using `.. notebooklite:: notebook_X.md` would convert `notebook_X.md` to `notebook_X_converted.ipynb` (or perhaps something like `notebook_X.md.jupyterlite`) and open it, which would perform both actions: resolving such an edge case, and keeping both notebooks intact in the documentation's source folder.

The docs build is failing because of this, but I would like to receive further direction before I proceed with this approach.

agriyakhetarpal:
Thanks – will do when this is reviewed and released!

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: The commit 04014268ce9a7ec0c1fa9a676b5c73c881979db6 resolves this, handling the collision by using a `_converted` suffix for the MyST notebook but leaves IPyNB files intact. For example, it can be tried out here: https://jupyterlite-sphinx--221.org.readthedocs.build/en/221/directives/voici.html for the IPyNB and the MD variants.

My observation regarding UX for this is that this change in the notebooks' filename(s) would not be noticeable for usage within the `Voici` and `NotebookLite` directives, since they do not showcase the filename at all when the kernel is spawned. The `JupyterLite` directive, however, does do so (regardless of whether we are in a new tab or within the same tab), but I think most users won't find it bothersome. Further, if one switches to the "Simple Interface" under "Appearance", the notebook occupies the entire screen and does not appear in a tab.

I wonder if there is a method to open notebooks under our `JupyterLite`'s new tab option with the Simple Interface directly, though, because it's neater that way... I'll follow up on https://github.com/jupyterlite/jupyterlite/issues/385.

agriyakhetarpal:
Hi @gabalafou, I've ensured that these Markdown/MyST notebooks can be stripped via 6a6d7a2a6bb5c85fa08f9b5a2b1818f11ac82adf, and added docs on how to use the stripping functionality with Markdown notebooks in 586742819f54c3edb0fa5f1a07555991975341c7 – as we were discussed yesterday. I have tested the changes that I have also upstreamed to a SciPy PR here: scipy/scipy#22042. Based on a self-review, things look alright to me to go ahead with.

agriyakhetarpal:
The case where `my_notebook.ipynb` and `my_notebook.md` both exist in the same folder has been handled, but I found yet another edge case based on these collisions, which is slightly trickier to resolve (but also much less likely to happen). If we have multiple files of the same name that exist across folders, we would have a problem. For example,

- `my_notebook.ipynb` exists in `folder1/`, `my_notebook.md` exists in `folder1/` and `my_notebook.ipynb` exists in `folder2/`

We would have a conflict again because all of the notebooks are going to be copied under the `contents/` directory. We should preserve the folder structure when copying to the `contents/` folder (i.e., create `contents/folder1/` and `contents/folder2/` and copy the notebooks in there) and indeed add UUID suffixes for the files to ensure that no collisions happen.

But, at the same time, it is safe to say that this case would have existed prior to this PR as well (my example is about two `.ipynb` files, and I am not sure if there are any examples of anyone having faced it in the wild (no bug reports so far, at least). Would it be okay to proceed with this PR as a known failure case and open a separate issue for it?

agriyakhetarpal:
The case where `my_notebook.ipynb` and `my_notebook.md` both exist in the same folder has been handled, but I found yet another edge case based on these collisions, which is slightly trickier to resolve (but also much less likely to happen). If we have multiple files of the same name that exist across folders, we would have a problem. For example,

- `my_notebook.ipynb` exists in `folder1/`, `my_notebook.md` exists in `folder1/` and `my_notebook.ipynb` exists in `folder2/`

We would have a conflict again because all of the notebooks are going to be copied under the `contents/` directory. We should preserve the folder structure when copying to the `contents/` folder (i.e., create `contents/folder1/` and `contents/folder2/` and copy the notebooks in there) and indeed add UUID suffixes for the files to ensure that no collisions happen.

However, it is safe to say that this case existed prior to this PR as well (my example concerns two `.ipynb` files). I am not sure if anyone has encountered it in the wild – at least, there have been no bug reports about it so far. Would it be okay to proceed with this PR as is, mark this behaviour as a known case of failure, and open a separate issue for it?

agriyakhetarpal:
Rebased on top of `main` to fix merge conflicts.

agriyakhetarpal:
Rebased on top of `main` to fix merge conflicts I recently introduced.

agriyakhetarpal:
> The case where `my_notebook.ipynb` and `my_notebook.md` both exist in the same folder has been handled, but I found yet another edge case based on these collisions, which is slightly trickier to resolve (but also much less likely to happen). If we have multiple files of the same name that exist across folders, we would have a problem. For example,
> 
> * `my_notebook.ipynb` exists in `folder1/`, `my_notebook.md` exists in `folder1/` and `my_notebook.ipynb` exists in `folder2/`
> 
> We would have a conflict again because all of the notebooks are going to be copied under the `contents/` directory. We should preserve the folder structure when copying to the `contents/` folder (i.e., create `contents/folder1/` and `contents/folder2/` and copy the notebooks in there) and indeed add UUID suffixes for the files to ensure that no collisions happen.
> 
> However, it is safe to say that this case existed prior to this PR as well (my example concerns two `.ipynb` files). I am not sure if anyone has encountered it in the wild – at least, there have been no bug reports about it so far. Would it be okay to proceed with this PR as is, mark this behaviour as a known case of failure, and open a separate issue for it?

This was discussed with @steppi today, and we agreed upon making a naming conflict a hard error, regardless of file extension. He suggested that I check if any projects have notebooks with duplicate names, so that we could correct them downstream as a safety measure before we release. I didn't find any for NumPy, SciPy, or scikit-learn; and neither has anyone reported an issue about this before – ever since `jupyterlite-sphinx` has had its inception. So, it is all the more the right time to conform to correct behaviour, and if someone raises this, we could just ask them to rename the duplicate notebook they have – at most, it would cost them a broken URL or few, which they can fix with redirects via RTD or through any other methods that we are not concerned about. Or, if we get a lot of bug reports about it, we could look into mangling the filenames with UUIDs and the like (which was my previous approach).

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: This conversation was resolved via https://github.com/jupyterlite/jupyterlite-sphinx/pull/221#issuecomment-2546726073.

agriyakhetarpal:
I've gone one step ahead in the check and also allowed having notebooks of the same filename stem—such as `my_notebook.ipynb` and `my_notebook.md`—to co-exist anywhere in the folder structure, but only under the condition where both of them are not referenced in our directives. This way, we are not going to deal with notebooks that are not our business – if, say, if only `my_notebook.ipynb` has been included in a `.. jupyterlite:: my_notebook.ipynb` directive somewhere in the docs by a user, it is only then that we process it further. 

Also, we perform this lookup with a set, so that operation happens in linear time anyway – I did think about caching it using a frozenset with paths serialised to strings and a simple LRU approach, but I didn't feel it was worth keeping because we are never going to expect notebooks in the order of thousands.

This is ready for review, now!

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -410,43 +484,58 @@ def run(self):
...
-                    for cell in nb.cells
-                    if "jupyterlite_sphinx_strip" not in cell.metadata.get("tags", [])
-                ]
-                nbformat.write(nb, notebooks_dir, version=4)
-
-            # If notebook_is_stripped is False, then copy the notebook(s) to notebooks_dir.
-            # If it is True, then they have already been copied to notebooks_dir by the
-            # nbformat.write() function above.
+            if notebook_path.suffix.lower() == ".md":
+                if self._should_convert_notebook(notebook_path, target_path):
```
steppi, line 513:
```suggestion
                if self._target_is_stale(notebook_path, target_path):
```

corresponds with an earlier suggestion. makes it more clear what's going on in my opinion

steppi (commented):
This is looking good Agriya. A couple small suggestions. Only real blocker I think is that the error message for when duplicate notebooks are found could use a little work. Let's discuss it offline when you have time.

agriyakhetarpal:
Albert and I discussed the error message offline and in consideration, reduced its size because of its verbosity; plus, it would be weird to see it as such from a user's perspective. It made more sense to make it more impersonal, highlighting the problem of conflicting notebook names at hand. Implemented in 81c0c09f5bfe162c503ef62d8f978725e3a0da7a.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -390,6 +391,79 @@ class _LiteDirective(SphinxDirective):
...
+            if colliding_files:
+                colliding_files.append(str(source_path))
+                raise ValueError(
+                    "Found multiple notebooks in the documentation sources marked for "
+                    f"inclusion with JupyterLite that would convert to '{target_ipynb}'.\n"
+                    f"The conflicting files are: {', '.join(colliding_files)}. \n"
+                    "Please rename them to avoid conflicts, as having multiple "
+                    "notebooks with the same name regardless of their file extension "
+                    "is not supported since it can lead to unexpected behaviours."
+                )
```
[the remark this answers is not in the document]
agriyakhetarpal: The error type was changed in d770338f657d0c83ddc0082a33a34ff487430d46 and the error message was replaced in 81c0c09f5bfe162c503ef62d8f978725e3a0da7a.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -390,6 +391,79 @@ class _LiteDirective(SphinxDirective):
...
+        some rudimentary timestamp-based caching."""
+        if not target_path.exists():
+            return True
+
+        return source_path.stat().st_mtime > target_path.stat().st_mtime
+
+    # TODO: Jupytext support many more formats for conversion, but we only
+    # consider Markdown and IPyNB for now. If we add more formats someday,
+    # we should also consider them here.
+    def _get_target_name(self, source_path: Path, notebooks_dir: Path) -> str:
```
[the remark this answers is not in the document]
agriyakhetarpal: Thanks, addressed in 574acae60e9f4fcba806ef6d017b3ef8384899fc.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -410,43 +484,58 @@ def run(self):
...
+            notebook_path = Path(notebook)
 
-            notebooks_dir = Path(self.env.app.srcdir) / CONTENT_DIR / notebook_name
+            self.env.jupyterlite_notebooks.add(str(notebook_path))
+
+            notebooks_dir = Path(self.env.app.srcdir) / CONTENT_DIR
+            os.makedirs(notebooks_dir, exist_ok=True)
+
+            target_name = self._get_target_name(notebook_path, notebooks_dir)
+            target_path = notebooks_dir / target_name
```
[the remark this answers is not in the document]
agriyakhetarpal: Thanks, addressed in 574acae60e9f4fcba806ef6d017b3ef8384899fc.

agriyakhetarpal:
With 29bb81015a67f34090834f9cf8b66d59846dc617, we don't keep going until we find all conflicts at once, and stop at the first one itself.

steppi:
Closing and reopening to investigate XPython kernel not being found in the readthedocs preview in https://github.com/jupyterlite/jupyterlite-sphinx/pull/228#issuecomment-2557874829

agriyakhetarpal:
🤔 I cannot reproduce it here either

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -390,6 +391,72 @@ class _LiteDirective(SphinxDirective):
...
+        target_ipynb = f"{target_stem}.ipynb"
+
+        # Only look for conflicts in source directories and among referenced notebooks.
+        # We do this to prevent conflicts with other files, say, in the "_contents/"
+        # directory as a result of a previous failed/interrupted build.
+        if source_path.parent != notebooks_dir:
+
+            # We only consider conflicts if notebooks are actually referenced in
+            # a directive, to prevent false posiitves from being raised.
+            if hasattr(self.env, "jupyterlite_notebooks"):
```
steppi, line 420: I think it should be fine for now, but the $O(n^2)$ algorithm used here might be painful for projects with a very large number of notebooks. We could do $O(n)$ if we store the basenames in a set. Just making a note of it in case this ends up being a bottleneck for someone.

steppi:
Thanks @agriyakhetarpal. The issue with the docs was transient. I'm still not sure what the problem was

PR #223, opened 2024-12-11
Add an option to open the Notebook UI and Voici apps in a new tab via the`NotebookLite` and `Voici` directives

agriyakhetarpal:
## Description

This PR implements #165, but for the Notebook UI and the Voici interface, which come to life via the `..notebooklite::` and `.. voici::` directives and notebooks can now be opened in a separate tab instead of within the documentation website. With it, we also receive the added benefit of a smaller button for both directives IFrames, instead of a larger-sized, framed 400px window element on the screen.

## Rationale

When running a notebook, the `JupyterLite` interface displays the entire JupyterLite Lab interface. The ability to use the Notebook UI in a new tab would be useful because it is more minimal and designed for those who do not need to display all the notebooks in the file browser drawer, need access to several other options, etc.

Unlike in JupyterLab, the "Simple Interface" has not yet been directly implemented in JupyterLite (please see jupyterlite/jupyterlite#385). Thus, this change will allow us to offer a reasonable substitute for it at the time, and bring parity between both notebook interface spawners.

The `Voici` directive can currently open notebook files if a notebook is not supplied to it. However, it, too, occupies a lot of space and one has to first load Voici and then open a notebook. By adding a "new tab" button, we have the provision to skip this step and open the notebook directly, and save space on the screen similar to the "-Lite" directives. This brings parity between all three major directives (the Replite console is not relevant here).

## Changes made

- Abstracted the "open in a new tab" functionality with a "base" class for the tabbed versions of both `JupyterLite` and `NotebookLite` directives
- Created a tabbed interface for the Voici directive
- Added docs for the new `:new_tab:` option for the `NotebookLite` directive and clarified the difference between both directive's "new tab" options
- Added docs for the new `:new_tab:` option for the `Voici` directive

## Additional context

After this change, the NotebookLite directive can be used in place of the JupyterLite directive for many of SciPy's interactive notebooks under `scipy.stats`: https://scipy.github.io/devdocs/tutorial/stats/hypothesis_tests.html, so that a notebook, when opened in a new tab, can occupy the full width on the screen (and distracting elements such as other notebooks in the UI and extra buttons can also be reduced).

cc: @melissawm, @steppi

### DIFF
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -204,7 +204,15 @@ class JupyterLiteIframe(_LiteIframe):
     notebooks_path = ""
 
 
-class JupyterLiteTab(_InTab):
+class BaseNotebookTab(_InTab):
+    """Base class for notebook tab implementations. We subclass this
+    to create more specific configurations around how tabs are rendered."""
+
+    lite_app = None
+    notebooks_path = None
+
+
+class JupyterLiteTab(BaseNotebookTab):
     """Appended to the doctree by the JupyterliteDirective directive
 
     Renders a button that opens a Notebook with JupyterLite in a new tab.
@@ -212,6 +220,16 @@ class JupyterLiteTab(_InTab):
 
     lite_app = "lab/"
     notebooks_path = ""
+
+
+class NotebookLiteTab(BaseNotebookTab):
+    """Appended to the doctree by the NotebookliteDirective directive
+
+    Renders a button that opens a Notebook with NotebookLite in a new tab.
+    """
+
+    lite_app = "tree/"
+    notebooks_path = "../notebooks/"
 
 
 class NotebookLiteIframe(_LiteIframe):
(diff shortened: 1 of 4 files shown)


agriyakhetarpal:
Initially, I planned for this improvement to be limited to just the `NotebookLite` directive, but the same thing for the `Voici` directive wasn't much trouble to implement either (please see the "Rationale" section in the PR description). We could restrict it to being just for the `NotebookLite` directive if needed.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -239,20 +272,57 @@ def __init__(
...
+            rawsource,
+            **attributes,
+        )
+
+    def html(self):
+        return (
+            '<button class="try_examples_button" '
+            f"onclick=\"window.open('{self.lab_src}')\">"
+            "Open with Voici</button>"
+        )
```
Carreau, line 323: Can't you use `<button target="_blank">` to open in a new tab w/o js ? it may need to be a `<a>` though.

Carreau: I see this is how other buttons are doing this; so I think this is fine for this PR, but we shoudl review other usages and try to make them not use JS for accessibility

Carreau (approved):

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: Yes, it sounds like a good idea. Though, we'll need to ensure the styling works as well, because we are using the `try_examples_button` class almost everywhere in the code. I'm unsure how much work would be needed. I'll open a new issue for this.

Side note: I'd also like to be able to modify the text within the buttons, such that instead of "Open with Voici", a user can say, "Start app" (on both a per-button and global basis).

agriyakhetarpal:
Rebased on top of `main` and checked all directives on the Read the Docs PR preview again, they are working as expected. Thanks for the review, @Carreau!

Issue #124, opened 2024-01-23
Provide a theme toggle plugin by default


martinRenou:
### Problem

As suggested in https://github.com/pydata/pydata-sphinx-theme/issues/745#issuecomment-1905560716

> @jtpio just linked this in another thread https://jupyterlite.readthedocs.io/en/stable/howto/configure/advanced/iframe.html
> 
> It showcases an example making a jupyterlite plugin allowing to set the theme from outside of the iframe.
> 
> We could provide such a plugin by default in jupyterlite-sphinx, allowing pydata-sphinx-theme to notify jupytelite iframes of a potential theme change.

We could no only support setting the theme using a theme name, but the message could probably also contain hardcoded css variable values as a dictionary.

agriyakhetarpal:
I think this is possible via the [`jupyterlab-theme-toggler`](https://github.com/jupyterlab-contrib/jupyterlab-topbar/tree/main/packages/theme-toggler-extension#jupyterlab-4x) extension now in the JupyterLite UI. But I don't think it provides a hook to set the theme from outside the iframe yet...?

jtpio:
From outside the IFrame there is the `theme` parameter, which can specified when embedding the IFrame: https://jupyterlite.readthedocs.io/en/stable/quickstart/embed-repl.html#themes

Then when it comes to changing the theme after the IFrame is displayed, we can indeed use the approach linked above, or a helper package like https://github.com/TileDB-Inc/jupyter-iframe-commands to execute arbitrary commands in the IFrame, from a host page.

PR #225, opened 2024-12-12
Allow the use of a custom `overrides.json` file for configuring JupyterLite at runtime

agriyakhetarpal:
## Description

This PR introduces a configuration option in `conf.py` called `jupyterlite_overrides`. This option modifies the command used to build JupyterLite internally within `jupyterlite-sphinx` to use a user-provided `overrides.json` file.

## Rationale

- It is currently possible to leave an `overrides.json` file in the `docs/` directory, similar to [`ipycanvas`'s documentation](https://github.com/jupyter-widgets-contrib/ipycanvas/blob/7226d36cc05e0f15b03af673510ed3e0f3cda424/docs/overrides.json), which will automatically be picked up by `jupyterlite-sphinx` (and thereby JupyterLite) when building the JupyterLite site.
- However, projects using `jupyterlite-sphinx` can benefit from the flexibility of using a custom override file with the docs, which is possible if one has access to the JupyterLite build command with ease.
This option is already available via https://jupyterlite-sphinx.readthedocs.io/en/latest/configuration.html#additional-cli-arguments-for-jupyter-lite-build, which was established via #169. Still, that approach has been directed mostly towards advanced users, and having this feature exposed through a configuration option similar to one for a build-time configuration file would make it easier to use and make it more visible with the rest of the options in the docs.

## Additional context

Please see https://github.com/jupyterlite/jupyterlite/issues/385#issuecomment-2536286348, which is where I got this idea. Some possible use cases for this can be:
- to be able to add download buttons for interactive notebooks within SciPy and PyWavelets that have been/are being established via https://github.com/scipy/scipy/pull/20303/ (cc: @melissawm) and https://github.com/PyWavelets/pywt/pull/741
- customise themes based on the surrounding theme of the page (dark or light – not sure how to switch it yet, but we do have an open issue about it: #124).

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -46,7 +46,7 @@ dependencies:
 
 ## JupyterLite configuration
 
-You can provide [custom configuration files](https://jupyterlite.readthedocs.io/en/latest/howto/configure/config_files.html)
+You can provide [custom configuration files](https://jupyterlite.readthedocs.io/en/stable/howto/configure/config_files.html)
 to your JupyterLite deployment for build-time configuration and settings overrides.
 
 The build-time configuration can be used to change the default settings for JupyterLite, such
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -49,6 +49,14 @@ dependencies:
 You can provide [custom configuration files](https://jupyterlite.readthedocs.io/en/latest/howto/configure/config_files.html)
 to your JupyterLite deployment for build-time configuration and settings overrides.
 
+The build-time configuration can be used to change the default settings for JupyterLite, such
+as changing which assets are included, the locations of the assets, which plugins are enabled,
+and more.
+
+The runtime configuration can be used to change the settings of the JupyterLite deployment
+after it has been built, such as changing the theme, the default kernel, the default language,
+and more.
+
 <!-- TODO: Run-time configuration via `jupyter-lite.json` not added yet here
 because I can't yet find a direct jupyter lite CLI mapping for that option -->
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -44,13 +44,19 @@ dependencies:
   - ipycanvas
 ```
 
-## JupyterLite config
+## JupyterLite configuration
 
-You can provide [custom configuration](https://jupyterlite.readthedocs.io/en/latest/howto/index.html#configuring-a-jupyterlite-deployment)
-to your JupyterLite deployment.
+You can provide [custom configuration files](https://jupyterlite.readthedocs.io/en/latest/howto/configure/config_files.html)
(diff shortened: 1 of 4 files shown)


agriyakhetarpal:
I've (temporarily) pushed bf5e616dbc2cdcb829345c08ab55993cff94db67, which enables the https://github.com/Quansight-Labs/jupyterlab-theme-christmas theme and a download button for notebooks within our documentation. It is working nicely and as expected, and the changes can be tried out in the PR preview: https://jupyterlite-sphinx--225.org.readthedocs.build/en/225/

A customised JupyterLite deployment within the the Sphinx documentation PR previews for jupyterlite-sphinx where JupyterLite can be seen running with the 'JupyterLab Christmas' theme. A download button is also available for the notebook.

dev-environment.yml
```diff
@@ -16,3 +16,5 @@ dependencies:
     - pip:
         - .
         - jupyterlite-xeus >=0.2.0a0
+        # Temporary for this PR
+        - jupyterlab_theme_christmas
```
Carreau, line 20: 👍

docs/configuration.md
```diff
@@ -44,13 +44,27 @@ dependencies:
   - ipycanvas
 ```
 
-## JupyterLite config
+## JupyterLite configuration
 
-You can provide [custom configuration](https://jupyterlite.readthedocs.io/en/latest/howto/index.html#configuring-a-jupyterlite-deployment)
-to your JupyterLite deployment.
+You can provide [custom configuration files](https://jupyterlite.readthedocs.io/en/latest/howto/configure/config_files.html)
```
Carreau, line 49: Side question; I know it was already in the readme but do we want to point to latest or stable ?

agriyakhetarpal: Good point, I think pointing to "stable" is better any day. Will change it

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -772,6 +789,7 @@ def setup(app):
 
     # Config options
     app.add_config_value("jupyterlite_config", None, rebuild="html")
+    app.add_config_value("jupyterlite_overrides", None, rebuild="html")
```
Carreau, line 792: 👍

Carreau (approved):

docs/configuration.md
agriyakhetarpal: Addressed in 9b241ce8b3f9e9de6361d5505e9e852318ba8505.

agriyakhetarpal:
Rebased on top of `main` and removed the Christmas theme I had added for experimentation. I can confirm that it continues to work with a local build. Thanks again for the review, @Carreau.

PR #227, opened 2024-12-13
Allow global and custom button text for the new-tabbed variants of the `JupyterLite`, `NotebookLite`, and the `Voici` directives

agriyakhetarpal:
## Description

This PR implements three global configuration options: `jupyterlite_button_text`, `notebooklite_button_text`, and `voici_button_text`; to customise the buttons for the tabbed variants of these directives – which were introduced via #165 and #223. This is similar to the `TryExamples` button text options, which have a similar pathway through a global value and per-directive overrides.

This allows users to make use of the extended ability to customise the button text as per their liking, say, "Run notebook"
in comparison to being restricted to the current "Open as a notebook".

Please note that this is valid only when `:new_tab:` is set – i.e., when we have a button present in the first place, because all three directives also have a `:prompt:` option for this purpose when they are displayed as iframes and not as buttons.

## Additional context

This has been implemented in continuation from https://github.com/jupyterlite/jupyterlite-sphinx/pull/223#discussion_r1883687821.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -67,6 +67,23 @@ jupyterlite_config = "jupyter_lite_config.json"
 jupyterlite_overrides = "overrides.json"
 ```
 
+# Setting default button texts for the `JupyterLite`, `NotebookLite`, and `Voici` directives
+
+When using the `:new_tab:` option in the `JupyterLite`, `NotebookLite`, and `Voici` directives,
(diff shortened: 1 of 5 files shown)


Carreau (approved):

agriyakhetarpal:
This option does not exist for the `Replite` directive because we don't have a new-tabbed version of it yet; I am on the fence about whether we need that or not? It would be relatively easy to implement.

PR #228, opened 2024-12-13
Add a new-tabbed variant for the `Replite` directive, and allow customisation of its button text

agriyakhetarpal:
## Description

This continues #165, #223, and #227 for the Replite directive. We do not subclass from the `BaseNotebookTab` class here since we need to construct a URL with quoted code that has been passed to the directive (similar to the `TryExamples` directive).

I was initially unsure about doing this as I mentioned in https://github.com/jupyterlite/jupyterlite-sphinx/pull/227#issuecomment-2541405294, but the implementation seems to be in line enough with that for other directives and felt straightforward to me. This PR shall bring feature parity for all four major directives, and all five directives can now have new-tabbed variants. As a plus, customisation is allowed for this directive's new-tab button text as well, either by a global `replite_new_tab_button_text` value or on a per-directive basis.

## Changes made

- A new `RepliteTab` class that inherits from `docutils.nodes.Element`. Here, we construct a URL for the REPL based on the input; and we return it when the `:new_tab:` option is enabled; otherwise, we return `RepliteIframe` instead.
- Documentation updated for the global `replite_new_tab_button_text` option for `conf.py`, and a sample directive added to showcase Replite opening in a new tab.
- The `:button_text:` option for the "-Lite" directives has been renamed to `:new_tab_button_text:` for clarity

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -82,7 +82,7 @@ replite_button_text = "My custom Replite button text"
 voici_button_text = "My custom Voici button text"
 ```
 
-You can override this text on a per-directive basis by passing the `:button_text:` option
+You can override this text on a per-directive basis by passing the `:new_tab_button_text:` option
 to the directive. Note that this is compatible only if `:new_tab:` is also provided.
 
 ## Strip particular tagged cells from IPython Notebooks
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -67,10 +67,10 @@ jupyterlite_config = "jupyter_lite_config.json"
 jupyterlite_overrides = "overrides.json"
 ```
 
-# Setting default button texts for the `JupyterLite`, `NotebookLite`, and `Voici` directives
+# Setting default button texts for the `JupyterLite`, `NotebookLite`, `Replite`, and `Voici` directives
 
-When using the `:new_tab:` option in the `JupyterLite`, `NotebookLite`, and `Voici` directives,
-the button text defaults to "Open as a notebook" and "Open with Voici", respectively.
+When using the `:new_tab:` option in the `JupyterLite`, `NotebookLite`, `Replite`, and `Voici` directives,
+the button text defaults to "Open as a notebook", "Open in a REPL", and "Open with Voici", respectively.
 
 You can optionally the button text on a global level for these directives by setting the
 following values in your `conf.py` file:
@@ -78,6 +78,7 @@ following values in your `conf.py` file:
 ```python
 jupyterlite_button_text = "My custom JupyterLite button text"
 notebooklite_button_text = "My custom NotebookLite button text"
+replite_button_text = "My custom Replite button text"
 voici_button_text = "My custom Voici button text"
 ```
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -76,10 +76,10 @@ You can optionally the button text on a global level for these directives by set
 following values in your `conf.py` file:
 
 ```python
-jupyterlite_button_text = "My custom JupyterLite button text"
+jupyterlite_new_tab_button_text = "My custom JupyterLite button text"
 notebooklite_button_text = "My custom NotebookLite button text"
-replite_button_text = "My custom Replite button text"
-voici_button_text = "My custom Voici button text"
+replite_new_tab_button_text = "My custom Replite button text"
+voici_new_tab_button_text = "My custom Voici button text"
 ```
 
 You can override this text on a per-directive basis by passing the `:new_tab_button_text:` option
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -77,7 +77,7 @@ following values in your `conf.py` file:
 
 ```python
 jupyterlite_new_tab_button_text = "My custom JupyterLite button text"
-notebooklite_button_text = "My custom NotebookLite button text"
+notebooklite_new_tab_button_text = "My custom NotebookLite button text"
 replite_new_tab_button_text = "My custom Replite button text"
 voici_new_tab_button_text = "My custom Voici button text"
 ```
--- a/docs/directives/replite.md
+++ b/docs/directives/replite.md
@@ -75,13 +75,13 @@ with the code pre-filled.
 ```
 
 When using this option, it is also possible to customise the button text, overriding the
-global value using an additional `:button_text:` parameter:
+global value using an additional `:new_tab_button_text:` parameter:
 
 ```rst
 .. replite::
    :kernel: xeus-python
    :new_tab: True
-   :button_text: My custom Replite button text
+   :new_tab_button_text: My custom Replite button text
 
    import matplotlib.pyplot as plt
    import numpy as np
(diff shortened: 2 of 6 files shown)


agriyakhetarpal:
I've opened yet another PR today where your review would be appreciated, @Carreau 😁 Thanks a lot!

docs/directives/replite.md
```diff
@@ -38,3 +38,75 @@ This directive takes extra options which are the same options as the `replite` p
    ax.plot(x, y)
    plt.show()
 ```
+
+If you use the `:new_tab:` option in the directive, the Replite console will be opened in a new browser tab
```
steppi, line 42: image

Something's going on in the docs here where the there's very little space between the replite iframe and this line of text.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -356,19 +418,51 @@ def run(self):
...
             os.path.join(self.env.app.srcdir, JUPYTERLITE_DIR),
             os.path.dirname(self.get_source_info()[0]),
         )
 
+        if new_tab:
+            directive_button_text = self.options.pop("button_text", None)
+            if directive_button_text is not None:
+                button_text = directive_button_text
+            else:
+                button_text = self.env.config.replite_button_text
```
steppi, line 437:
```suggestion
        if new_tab:
            directive_button_text = self.options.pop("button_text", None)
            if directive_button_text is not None:
                button_text = directive_button_text
            else:
                button_text = self.env.config.replite_button_text
            return [
                RepliteTab(
                    prefix=prefix,
                    width=width,
                    height=height,
                    prompt=prompt,
                    prompt_color=prompt_color,
                    content=content,
                    search_params=search_params,
                    lite_options=self.options,
                    button_text=button_text,
                )
            ]
```

It makes sense to put this here to avoid having to branch on `new_tab` again.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -356,19 +418,51 @@ def run(self):
...
+        elif "button_text" in self.options:
+            raise ValueError(
+                "'button_text' is only valid if 'new_tab' is True. To modify the prompt text, use 'prompt' and 'prompt_color'."
+            )
```
steppi, line 441: Can we call this `"new_tab_button_text"` and just ignore it in `RepliteIframe`? Establishing this convention will also help us with `TryExamples` where the new tab button text isn't even configurable, but `button_text` is already used for the Try it button.

I saw you've established this same pattern elsewhere, in this file, but it looks like it hasn't made it into a release yet, so there's still time to change.

No one would be confused that `"new_tab_button_text"` doesn't control the prompt, so I think this would make things more clear. Also, in general, having options which are only allowed conditional on the value of other options is likely to create confusion.

steppi (changes requested):
Thanks Agriya. This ones looking good too. A suggestion about `"button_text"` and a couple minor suggestions.

docs/directives/replite.md
agriyakhetarpal: Yeah, I've noticed this problem before – I think this is the case for the NotebookLite and JupyterLite iframes as well. Should I attempt to fix it here or add a follow-up?

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: Thanks – addressed in 2f96d85b6d259563ec6118dc0808f7ec234f356e.

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: Yes, makes sense to modify the pattern everywhere, and +1 on not aiming to confuse people – I had the same thought!

I've renamed the option to `"new_tab_button_text"` for all three directives where applicable in 5d73efd0c653e1def463a84daaffeba48c5f2c37 and 874dbfd4cc0092f42eec83ff42e656222125973e, though I kept the variable name being passed as `button_text` internally, because the name of the class (`RepliteTab`) and its docstring (`"opens a REPL with JupyterLite in a new tab"`) inherently imply that the button text concerns a new tab.

Should I also rename the corresponding configuration options `jupyterlite_button_text` et al to `jupyterlite_new_tab_button_text`? I do feel that would be a bit long, but at the same time, I do get that it might instigate confusion on the difference between the prompt text and the button text (if one doesn't know what the iframes/buttons are supposed to represent visually).

docs/directives/replite.md
steppi: Follow up is fine

jupyterlite_sphinx/jupyterlite_sphinx.py
steppi: Cool. Thanks Agriya. If `jupyterlite_button_text` and friends haven't made it into a release yet, I like `jupyterlite_new_tab_button_text` better, even though a little long. In the context  of writing a config file, it's not that much typing, and makes things more clear.

docs/directives/replite.md
agriyakhetarpal: Opened a new issue about it at #232, thanks!

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: Sure – done in af260f6c0cb5bfbd74878432016f90629635feec. Yes, these options haven't made it to a release yet. Thanks for the correction.

agriyakhetarpal: Sure – done in af260f6c0cb5bfbd74878432016f90629635feec. Yes, these options haven't made it to a release yet. Thanks for the correction!

agriyakhetarpal:
Thanks for the review! I think I have addressed all points.

docs/configuration.md
```diff
@@ -67,21 +67,22 @@ jupyterlite_config = "jupyter_lite_config.json"
...
+When using the `:new_tab:` option in the `JupyterLite`, `NotebookLite`, `Replite`, and `Voici` directives,
+the button text defaults to "Open as a notebook", "Open in a REPL", and "Open with Voici", respectively.
 
 You can optionally the button text on a global level for these directives by setting the
 following values in your `conf.py` file:
 
 ```python
-jupyterlite_button_text = "My custom JupyterLite button text"
+jupyterlite_new_tab_button_text = "My custom JupyterLite button text"
 notebooklite_button_text = "My custom NotebookLite button text"
```
steppi, line 80:
```suggestion
notebooklite_new_tab_button_text = "My custom NotebookLite button text"
```

Should this be changed too?

agriyakhetarpal: Oops, thanks! Addressed in 6a28fa4f6c5fe1a989fd0d47fdf1e714d8d0c8a8.

steppi:
For some reason the XPython kernel isn't being found in the readthedocs preview at https://jupyterlite-sphinx--228.org.readthedocs.build/en/228/. Closing and reopening to retrigger CI to see if it's a transient issue.

agriyakhetarpal:
That's a bit strange, I tried it on the PR preview (cleared cookies and cache as well) before you reran the build, and it seems to load fine for me:

image

steppi:
It's strange yeah. I closed and reopened the markdown PR and the kernel is still found for me in the preview.

steppi (approved):
Looks good. Thanks @agriyakhetarpal. The issue with the docs was transient. I just didn't want to merge without trying out the docs.

PR #229, opened 2024-12-13
Bump OS image and dependencies for Read the Docs build config

agriyakhetarpal:
## Description

Our Read the Docs configuration is slightly outdated by now, so I've updated the following:
- the OS version from Ubuntu 20.04 to Ubuntu 22.04
- the Mambaforge version from 4.10 to 23.11

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,7 @@
 version: 2
 
 build:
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
The RTD build is working without any issues and isn't related to a bug or a user-facing feature; self-merging. Thanks!

PR #231, opened 2024-12-16
Use more accessible visual elements instead of buttons

### LINKED ISSUE #226: Review the button's onclick window.open for replacement by target=_blank
### Problem

A number of buttons use an onclik window.open to open in a new window, while `<a target='_blank'>` exist for this purpose.

This is not optimal as a hover on link with target-blank tell you what it does, but not if there is an onclick handler; and create accessibility issues.

### Proposed Solution

use `<a>`s ans style them like the buttons

### Additional context

https://github.com/jupyterlite/jupyterlite-sphinx/pull/223/files#r1883631155)

agriyakhetarpal:
This PR intends to close #226, and is currently a work in progress. I shall update the description soon once I have the buttons working the same as they are outside this PR.

### DIFF
--- a/docs/_static/try_examples.css
+++ b/docs/_static/try_examples.css
@@ -4,6 +4,9 @@
 }
 
 .try_examples_button {
+  text-decoration: none;
+  display: inline-block;
+  cursor: pointer;
   background-color: var(--jupyter-light-primary);
   border: none;
   padding: 5px 10px;
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -143,9 +143,10 @@ class _InTab(Element):
 
     def html(self):
         return (
-            '<button class="try_examples_button" '
-            f"onclick=\"window.open('{self.lab_src}')\">"
-            f"{self.button_text}</button>"
+            f'<a href="{self.lab_src}" '
+            'target="_blank" rel="noopener noreferrer" '
+            'class="try_examples_button" style="text-decoration: none; display: inline-block; color: inherit;">'
+            f"{self.button_text}</a>"
         )
 
 
@@ -322,9 +323,10 @@ class VoiciTab(Element):
 
     def html(self):
         return (
-            '<button class="try_examples_button" '
-            f"onclick=\"window.open('{self.lab_src}')\">"
-            f"{self.button_text}</button>"
+            f'<a href="{self.lab_src}" '
+            'target="_blank" rel="noopener noreferrer" '
+            'class="try_examples_button" style="text-decoration: none; display: inline-block; color: inherit;">'
+            f"{self.button_text}</a>"
         )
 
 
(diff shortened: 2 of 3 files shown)


Carreau:
Thanks for working on this !

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -675,17 +677,23 @@ def run(self):
 
         # Button with the onclick event to swap embedded notebook back to examples.
         go_back_button_html = (
-            '<button class="try_examples_button" '
-            f"onclick=\"window.tryExamplesHideIframe('{examples_div_id}',"
-            f"'{iframe_parent_div_id}')\">"
-            "Go Back</button>"
```
gabalafou, line 681: I see no reason to change this element. It really is a button; it's not link, it's **not** taking you to a new URL.

Generally speaking, you should only use the `role` attribute as a very last resort, in cases where you have constraints that prevent you from using the HTML tag that has the role you need. (For example, imagine you want to fix the semantics of the markup output by a tool, such as Sphinx, that's been around a long time. Imagine that there are a lot of downstream styles and code that have been built around that specific markup structure. Then you cannot safely change an `a`-tag to a `button`-tag without breaking a lot of downstream projects. So if you wanted to fix the semantics, your only safe option would be to use the `role` attribute.)

docs/_static/try_examples.css
```diff
@@ -4,6 +4,9 @@
 }
 
 .try_examples_button {
+  text-decoration: none;
+  display: inline-block;
+  cursor: pointer;
```
gabalafou, line 9: Why is this CSS needed?

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -322,9 +323,10 @@ def __init__(
 
     def html(self):
         return (
-            '<button class="try_examples_button" '
-            f"onclick=\"window.open('{self.lab_src}')\">"
-            f"{self.button_text}</button>"
+            f'<a href="{self.lab_src}" '
+            'target="_blank" rel="noopener noreferrer" '
```
gabalafou, line 327: I'm not entirely sure why you need `noopener` or `noreferrer`.

Generally speaking `noreferrer` is used if you're building something like a discussion forum. It removes the incentive for people to post spam links. Noreferrer says to Google something like: as the webmaster of this site, I disavow this link, it could be spam, please don't penalize my search engine rankings for including it on my site.

The general idea behind both `noopener` and `noreferrer` is that you don't know where you're linking to. You're linking to third-party content that you're concerned could potentially abuse your site or visitor via the link. As far as I understand, that's not the same case here because the try_examples link is created from content that you own and produce.

In any case, this is a change that should be in a separate pull request because it's not really relevant to the other changes in this pull request, and it changes the existing behavior.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -143,9 +143,10 @@ def __init__(
...
+            'class="try_examples_button" style="text-decoration: none; display: inline-block; color: inherit;">'
```
gabalafou, line 148: Aren't these styles duplicates of the stylesheet? I think styling them via `a.try_examples_button` should be sufficient.

gabalafou (commented):
I love seeing an accessibility-related pull request! :sunflower:

So I couldn't resist leaving a few comments, even though this is still in draft mode. :smile:

PR #233, opened 2024-12-20
Switch to Miniforge and drop Mambaforge

agriyakhetarpal:
## Description

CI runs are failing on the `main` branch and on recent PRs due to problems with setting `miniconda` up, this PR should fix the problem.

More descriptions here: https://conda-forge.org/news/2024/07/29/sunsetting-mambaforge/

## Changes made

- Bumped to `actions/checkout@v4` and `setup/miniconda@v3`
- Switched to Miniforge since the installer is identical

## Additional information

I feel adding a Dependabot file for GitHub Actions, and pinning the actions to their hashes would make sense as well, even if we set the update schedule to monthly.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,7 +32,7 @@ jobs:
         activate-environment: jupyterlite-sphinx
         environment-file: dev-environment.yml
         python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
This is a maintenance PR and not user-facing, so I'll go ahead and merge, thanks!

PR #249, opened 2025-01-10
Add `ConfigLoader` with throttled logging and `try_examples.json` config caching

agriyakhetarpal:
Closes #240

PR #250, opened 2025-01-10
Better mobile device detection for interactive examples buttons

### LINKED ISSUE #140: Hide button, or warn about high data download on mobile.

agriyakhetarpal:
## Description

> [!TIP]
> This PR builds on top of changes from #249 in order to ease merge conflicts. Please view the changes in commit f1bb4eaae483788012050b28760d7c3f8834fc58 for a better diff in the meantime.

This PR closes #140; it moves the logic from the CSS styles for disabling the interactive buttons and uses a better heuristic via our JavaScript code; i.e., the `navigator.userAgent`'s value to check what sort of device a user is on. As a fallback, the screen size check has also been kept. A resize handler with 250ms debouncing has also been added here to check the screen size if the buttons haven't been disabled already.

This keeps them enabled on devices like iPads and other tablets.

cc: @gabalafou

### DIFF
--- a/jupyterlite_sphinx/jupyterlite_sphinx.css
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.css
@@ -67,12 +67,3 @@
     transform: rotate(1turn);
   }
 }
-
-/* we do not want the button to show on smaller screens (phones), as clicking
- * can download a lot of data. 480px is a commonly used breakpoint to identify if a device is a smartphone.  */
-
-@media (max-width: 480px), (max-height: 480px) {
-  div.try_examples_button_container {
-    display: none;
-  }
-}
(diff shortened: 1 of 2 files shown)


agriyakhetarpal:
I rebased and updated the changes to work with the caching of the `try_examples.json` file, so that the mobile device detection is cached as well, and the checks for the resize of the page comply.

agriyakhetarpal:
I rebased and updated the changes to work with the caching of the `try_examples.json` file, so that the mobile device detection is also cached and the checks for the page's resize comply. Thanks for the review, @Carreau!

PR #251, opened 2025-01-10
Add the "See Also" section to the `TryExamples` directive's trailing section ignore patterns

agriyakhetarpal:
## Description

This PR is a bug fix for a problem I noticed in sympy/sympy#27419. [SymPy uses a vendored version of `numpydoc`](https://github.com/sympy/sympy/blob/3c33a8283cd02e2e012b57435bd6212ed915a0bc/doc/ext/numpydoc.py) that was added twelve years back based on https://github.com/numpy/numpydoc; both have diverged in implementation and complexity since then, and SymPy's `numpydoc` doesn't enforce a particular order for each section, unlike the current `numpydoc`, which can automatically order/reorder sections. This causes the following behaviour:

<details>
<summary>Tap to show "Before"</summary>

<br>

The image shows the documentation page for SymPy 1.14.dev's number theory module, specifically documenting the sympy.ntheory.factor_.multiplicity(p, n) function. The page features a green snake-wrapped-S logo and a dark green sidebar with navigation sections. The main content includes function documentation and an interactive JupyterLite notebook example showing function usage. Notably, the notebook contains an extra erroneous or rather misplaced 'seealso' section with :obj:~.trailing'` appended at the bottom, which appears to be a documentation artifact that shouldn't be part of the interactive example. The page also includes proper 'See also' references at the top linking to 'factorint' and 'smoothness' functions, and a right sidebar showing the 'Ntheory Class Reference' and related functions.

<br>

</details>

where the See Also section percolates into the notebook. With this fix in place applied, we have

<details>
<summary>Tap to show "After"</summary>

<br>

The image shows the documentation page for SymPy 1.14.dev's number theory module, specifically documenting the sympy.ntheory.factor_.multiplicity(p, n) function. The page features a green snake-wrapped-S logo and a dark green sidebar with navigation sections. The main content includes function documentation and an interactive JupyterLite notebook example showing function usage. Unlike the previous version noted above, this notebook correctly does not contain the erroneous 'seealso' section at the bottom, showing the proper behaviour. The page maintains its standard 'See also' references at the top linking to 'factorint' and 'smoothness' functions, and includes a right sidebar showing the 'Ntheory Class Reference' and related functions.

<br>

</details>

where the section no longer appears in the notebook, and we stop at the end of the Examples section, as intended.

## Additional context

It would be nice to update the SymPy documentation infrastructure to use the latest `numpydoc` so that the documentation is consistent with that for other packages. However, considering that its custom `numpydoc` has had multiple stylistic deviations by now, I've left it to be a follow-up task for me after the PR linked above gets merged. So, supporting SymPy in a special case in this manner doesn't sound too bad to me, considering that this change is trivial.

### DIFF
--- a/jupyterlite_sphinx/_try_examples.py
+++ b/jupyterlite_sphinx/_try_examples.py
@@ -303,6 +303,28 @@ _non_example_docstring_section_headers = (
     "Yields",
 )
 
+# Sourced from docutils.parsers.rst.directives._directive_registry.keys()
+# and from https://docutils.sourceforge.io/docs/ref/rst/directives.html
+# We hardcode a subset of this list as there is no public API in docutils
+# or Sphinx to get this information directly.
+# See https://docutils.sourceforge.io/docs/ref/rst/directives.html
+#
+# This subset is only a list of sensible defaults and widely used directives
+# based on our discretion that can signify the start of a new section in a
+# docstring and is not meant to be exhaustive.
+_irrelevant_directives = [
+    "class",
+    "contents",
+    "epigraph",
+    "footer",
+    "header",
+    "highlights",
+    "parsed-literal",
+    "pull-quote",
+    "seealso",
+    "sidebar",
+    "topic",
+]
 
 _examples_start_pattern = re.compile(r".. (rubric|admonition):: Examples")
 _next_section_pattern = re.compile(
@@ -315,10 +337,10 @@ _next_section_pattern = re.compile(
         + [r"\!\! processed by numpydoc \!\!"]
         # Attributes section sometimes has no directive.
         + [r":Attributes:"]
-        # The See Also section shows up for old numpydoc versions such as that
-        # of SymPy where it is accounted for as an admonition and not a rubric,
-        # therefore we check for it as a special case for now.
-        + [r"\.\. seealso::"]
+        # Directives that can start a new section in a docstring and are
+        # not handled by numpydoc in terms of reordering. Noticed in SymPy
+        # as it does not use modern numpydoc at the time of writing.
+        + [rf"\.\. ({directive})::" for directive in _irrelevant_directives]
     )
 )
 
--- a/jupyterlite_sphinx/_try_examples.py
+++ b/jupyterlite_sphinx/_try_examples.py
@@ -315,9 +315,9 @@ _next_section_pattern = re.compile(
         + [r"\!\! processed by numpydoc \!\!"]
         # Attributes section sometimes has no directive.
         + [r":Attributes:"]
-        # The See Also section shows up for old numpydoc versions where ordering
-        # is neither guaranteed nor enforced; such as SymPy, therefore we check
-        # for it as a special case.
+        # The See Also section shows up for old numpydoc versions such as that
+        # of SymPy where it is accounted for as an admonition and not a rubric,
+        # therefore we check for it as a special case for now.
         + [r"\.\. seealso::"]
     )
 )
--- a/jupyterlite_sphinx/_try_examples.py
+++ b/jupyterlite_sphinx/_try_examples.py
@@ -315,9 +315,12 @@ _next_section_pattern = re.compile(
         + [r"\!\! processed by numpydoc \!\!"]
         # Attributes section sometimes has no directive.
         + [r":Attributes:"]
-        # The See Also section shows up for old numpydoc versions such as that
-        # of SymPy where it is accounted for as an admonition and not a rubric,
-        # therefore we check for it as a special case for now.
+        # The See Also section is also a heading that needs to be escaped. This
+        # showed up in the SymPy documentation which used an older version of
+        # numpydoc that did not enforce ordering, and modern numpydoc did, hence
+        # it was not noticed. For future reference, see:
+        # https://github.com/jupyterlite/jupyterlite-sphinx/pull/251
+        # https://github.com/sympy/sympy/pull/27419
         + [r"\.\. seealso::"]
     )
 )
--- a/jupyterlite_sphinx/_try_examples.py
+++ b/jupyterlite_sphinx/_try_examples.py
@@ -315,6 +315,10 @@ _next_section_pattern = re.compile(
         + [r"\!\! processed by numpydoc \!\!"]
         # Attributes section sometimes has no directive.
         + [r":Attributes:"]
+        # The See Also section shows up for old numpydoc versions where ordering
+        # is neither guaranteed nor enforced; such as SymPy, therefore we check
+        # for it as a special case.
+        + [r".. seealso::"]
     )
 )
 
--- a/jupyterlite_sphinx/_try_examples.py
+++ b/jupyterlite_sphinx/_try_examples.py
@@ -303,28 +303,6 @@ _non_example_docstring_section_headers = (
     "Yields",
 )
 
-# Sourced from docutils.parsers.rst.directives._directive_registry.keys()
-# and from https://docutils.sourceforge.io/docs/ref/rst/directives.html
-# We hardcode a subset of this list as there is no public API in docutils
-# or Sphinx to get this information directly.
-# See https://docutils.sourceforge.io/docs/ref/rst/directives.html
-#
-# This subset is only a list of sensible defaults and widely used directives
-# based on our discretion that can signify the start of a new section in a
-# docstring and is not meant to be exhaustive.
-_irrelevant_directives = [
-    "class",
-    "contents",
-    "epigraph",
-    "footer",
-    "header",
-    "highlights",
-    "parsed-literal",
-    "pull-quote",
-    "seealso",
-    "sidebar",
-    "topic",
-]
 
 _examples_start_pattern = re.compile(r".. (rubric|admonition):: Examples")
 _next_section_pattern = re.compile(
@@ -337,10 +315,10 @@ _next_section_pattern = re.compile(
         + [r"\!\! processed by numpydoc \!\!"]
         # Attributes section sometimes has no directive.
         + [r":Attributes:"]
-        # Directives that can start a new section in a docstring and are
-        # not handled by numpydoc in terms of reordering. Noticed in SymPy
-        # as it does not use modern numpydoc at the time of writing.
-        + [rf"\.\. ({directive})::" for directive in _irrelevant_directives]
+        # The See Also section shows up for old numpydoc versions such as that
+        # of SymPy where it is accounted for as an admonition and not a rubric,
+        # therefore we check for it as a special case for now.
+        + [r"\.\. seealso::"]
     )
 )
 
--- a/jupyterlite_sphinx/_try_examples.py
+++ b/jupyterlite_sphinx/_try_examples.py
@@ -308,7 +308,7 @@ _examples_start_pattern = re.compile(r".. (rubric|admonition):: Examples")
 _next_section_pattern = re.compile(
     "|".join(
         [
-            rf".. (rubric|admonition)::\s*{header}"
+            rf"\.\. (rubric|admonition)::\s*{header}"
             for header in _non_example_docstring_section_headers
         ]
         # If examples section is last, processed by numpydoc may appear at end.
@@ -318,7 +318,7 @@ _next_section_pattern = re.compile(
         # The See Also section shows up for old numpydoc versions where ordering
         # is neither guaranteed nor enforced; such as SymPy, therefore we check
         # for it as a special case.
-        + [r".. seealso::"]
+        + [r"\.\. seealso::"]
     )
 )
 
--- a/jupyterlite_sphinx/_try_examples.py
+++ b/jupyterlite_sphinx/_try_examples.py
@@ -315,12 +315,8 @@ _next_section_pattern = re.compile(
         + [r"\!\! processed by numpydoc \!\!"]
         # Attributes section sometimes has no directive.
         + [r":Attributes:"]
-        # The See Also section is also a heading that needs to be escaped. This
-        # showed up in the SymPy documentation which used an older version of
-        # numpydoc that did not enforce ordering, and modern numpydoc did, hence
-        # it was not noticed. For future reference, see:
-        # https://github.com/jupyterlite/jupyterlite-sphinx/pull/251
-        # https://github.com/sympy/sympy/pull/27419
+        # See Also sections are mapped to Sphinx's `.. seealso::` directive,
+        # not admonitions or rubrics.
         + [r"\.\. seealso::"]
     )
 )


jupyterlite_sphinx/_try_examples.py
```diff
@@ -315,6 +315,10 @@ def _process_literal_blocks(md_text):
         + [r"\!\! processed by numpydoc \!\!"]
         # Attributes section sometimes has no directive.
         + [r":Attributes:"]
+        # The See Also section shows up for old numpydoc versions where ordering
+        # is neither guaranteed nor enforced; such as SymPy, therefore we check
+        # for it as a special case.
```
steppi, line 320: `"See Also"` is included in `_non_example_docstring_section_headers`, so I don't think the issue can hinge just on the ability of "See Also" sections to appear after an "Examples" sections in older numpydocs versions. It looks like in SymPy, after numpydoc preprocessing, the line for the See Also section is `".. seealso::"`, but the existing logic expects it to look like `".. rubric:: See Also`". I think the real issue is that the older numpydoc version uses a different format for the directives that get inserted during processing. It just so happens that SymPy has a `"See Also"` section after each `"Examples"` section, but I think any section appearing afterwards would fail to be detected.

If we want to support older numpydoc versions, then I think the principled solution would then be to either:

1. Add the format used by older numpydoc versions to `_next_section_pattern`
or
2. Detect the numpydoc version at runtime, and have `_next_section_start_pattern` depend on that version

jupyterlite_sphinx/_try_examples.py
```diff
@@ -315,6 +315,10 @@ def _process_literal_blocks(md_text):
...
+        + [r".. seealso::"]
```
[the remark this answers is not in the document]
steppi: ha, but it looks like i made the same mistake in line 311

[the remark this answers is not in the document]
agriyakhetarpal: Thanks for the catch – addressed both lines in 10abc82a39e8975727fb5ef2ddb77a400158c534

jupyterlite_sphinx/_try_examples.py
agriyakhetarpal: > It just so happens that SymPy has a `"See Also"` section after each `"Examples"` section, but I think any section appearing afterwards would fail to be detected.

Ah, you're right – thanks for pointing that out! I think it would be nice to support older `numpydoc` versions in whatever manner we can, but I think option 2) is not possible with SymPy's `numpydoc` at the moment, as it is present as a custom Sphinx extension in `docs/sphinxext/` across two files, and not as an installable package fetched from PyPI or a separate location. So we are left with 1), which I've done here.

agriyakhetarpal: Actually, the See Also section seems to be a case that has been handled specially: https://github.com/sympy/sympy/blob/3c33a8283cd02e2e012b57435bd6212ed915a0bc/doc/ext/docscrape_sphinx.py#L155-L161

And it doesn't seem to be the case for the "Notes", "Examples", or "References" sections: https://github.com/sympy/sympy/blob/3c33a8283cd02e2e012b57435bd6212ed915a0bc/doc/ext/docscrape.py#L424, which are not included in Try Ex

I haven't dived into the code in detail, but I infer that the difference is because the "See Also" section is an admonition in SymPy's `numpydoc`, and not a rubric. I'd suggest that we stick with the current approach for now, and add any other cases here if someone notices them, until I can migrate SymPy to the upstream `numpydoc` and remove their implementation (that is, I'm short of ideas, unless you have some).

steppi: If this is working, it's fine to do this hack to get SymPy working specifically, but the comment is incorrect and should be updated. It's not about the ordering but because of the form the directive takes in sympy's version of numpydocs.

agriyakhetarpal: Yes, makes sense to me. I hope 5f187f3272fbc0404d4ba437adcfc5e5df1205f3 addresses it!

jupyterlite_sphinx/_try_examples.py
```diff
@@ -308,13 +308,17 @@ def _process_literal_blocks(md_text):
...
+            rf"\.\. (rubric|admonition)::\s*{header}"
             for header in _non_example_docstring_section_headers
         ]
         # If examples section is last, processed by numpydoc may appear at end.
         + [r"\!\! processed by numpydoc \!\!"]
         # Attributes section sometimes has no directive.
         + [r":Attributes:"]
+        # The See Also section shows up for old numpydoc versions such as that
+        # of SymPy where it is accounted for as an admonition and not a rubric,
+        # therefore we check for it as a special case for now.
```
steppi, line 320: This is again not quite right. If you check line 311, it also checks for admonitions. The regex `rf"\.\. (rubric|admonition)::\s*{header}` will match `..` followed by one of `rubric` or `admonition`, followed by any number of whitespace, followed by one of the headers in `_non_example_docstring_headers`. 

But in at least SymPy, the See Also section doesn't get a directive like `.. admonition:: See Also` but just `.. seealso:`. What I'm curious about is, for current versions of numpydocs, does a See Also section get the directive `.. admonition:: See Also` like I assumed, or was I incorrect about this? 

Since the above logic does check for See Also sections, it can't be relevant that the See Also section shows up for old numpydoc versions. 
```suggestion
        # SymPy uses a custom numpydoc version which has a different pattern for
        #  the generated directives. Rather than investigating further, take advantage
        # of See Also being the only section appearing after Examples in SymPy and
        # hardcode the directive that appears in SymPy. This can be removed after
        # SymPy is updated to use regular numpydoc.
```

steppi: At some point someone should sit down and figure out what the generated directives can actually look like for each section header in mainline numpydoc.

agriyakhetarpal: I spent some time debugging this and while I couldn't get a definite answer about what happens where in the code, here's what I have realised –

There are two things at play here:
1. There is a [a `.. seealso` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-seealso) that's native to Sphinx, and [a "See Also" section](https://numpydoc.readthedocs.io/en/stable/format.html#see-also) that upstream `numpydoc` can interpret and convert to a rubric/admonition, which we correctly parse and escape with TryExamples' `_next_section_pattern` if we retain compatibility with upstream `numpydoc`.
2. However, SymPy's `numpydoc` is dated and can't do this for SymPy's documentation, and thus we don't get to catch it.

Therefore, a "See Also" section in the documentation for SciPy for an example, will caught in `numpydoc`, and this problem isn't specific to just `.. seealso::`, all native Sphinx-based directives won't get caught. It's just that this hasn't been reported as a bug before because of the following reasons:

A. most projects don't put anything after "Examples", and only SymPy that we know of (so far) has done so
B. Upstream `numpydoc` does indeed reorder the sections to some extent (which was my original comment a few commits ago), so even if a project does add a section, it gets caught by the pattern or re-ordered before it's rendered in the HTML

As a confirmation for point A, I tried another Sphinx directive from the page I linked above, such as [`.. versionadded::`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded), and here we go:

<details>
<summary>SymPy documentation screenshot with erroneous versionadded directive </summary>

<br>

image

</details>

The same experiment with NumPy confirms point B. For example, the docstring below:

<details><summary>NumPy's exceptions.AxisError docstring (modified)</summary>
<p>

<br>

```python
"""
    Examples
    --------
    >>> import numpy as np
    >>> array_1d = np.arange(10)
    >>> np.cumsum(array_1d, axis=1)
    Traceback (most recent call last):
      ...
    numpy.exceptions.AxisError: axis 1 is out of bounds for array of dimension 1

    Negative axes are preserved:

    >>> np.cumsum(array_1d, axis=-2)
    Traceback (most recent call last):
      ...
    numpy.exceptions.AxisError: axis -2 is out of bounds for array of dimension 1

    The class constructor generally takes the axis and arrays'
    dimensionality as arguments:

    >>> print(np.exceptions.AxisError(2, 1, msg_prefix='error'))
    error: axis 2 is out of bounds for array of dimension 1

    Alternatively, a custom exception message can be passed:

    >>> print(np.exceptions.AxisError('Custom error message'))
    Custom error message

    See Also
    --------

    numpy.core._exceptions._AxisError : The internal version of this exception.

    Attributes
    ----------
    axis : int, optional
        The out of bounds axis or ``None`` if a custom exception
        message was provided. This should be the axis as passed by
        the user, before any normalization to resolve negative indices.

        .. versionadded:: 1.22
    ndim : int, optional
        The number of array dimensions or ``None`` if a custom exception
        message was provided.

        .. versionadded:: 1.22
"""
```

</p>
</details> 


where the "See Also" section is after the "Examples" section – where we could assume that it would be ordered in the same manner. However, generating the NumPy docs, it shows up before the examples:

<details><summary>Screenshot 1 from NumPy documentation</summary>
<p>

image

</p>
</details> 

But, if you were to add the `.. versionadded::` directive for NumPy, it is neither reordered nor removed. Added a `.. versionchanged:: Agriya debugging` directive after the Examples section, and this is what we get:

<details><summary>NumPy documentation screenshots 2 and 3 with a native Sphinx-based directive after the Examples section</summary>
<p>

<br>

The HTML

image

and the interactive notebook

image

</p>
</details>

agriyakhetarpal: We'll need to add almost all of the directives listed under https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#paragraph-level-markup to be able to restrict them from percolating into the notebooks; I don't think it makes sense to leave it with just `.. seealso` especially if others like `..versionadded` are also included with both SymPy's `numpydoc` and the upstream `numpydoc`.

I used

```python
from docutils.parsers.rst import directives
all_directives = list(directives._directive_registry.keys())

Comment (agriyakhetarpal) [reply jupyterlite_sphinx/_try_examples.py:320]: We'll need to add almost all of the directives listed under https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#paragraph-level-markup to be able to restrict them from percolating into the notebooks; I don't think it makes sense to leave it with just `.. seealso` especially if others like `..versionadded` are also included with both SymPy's `numpydoc` and the upstream `numpydoc`.

I used this command

```python
from docutils.parsers.rst import directives
all_directives = list(directives._directive_registry.keys())
```

to get a list:

```python
['attention',
 'caution',
 'code',
 'danger',
 'error',
 'important',
 'note',
 'tip',
 'hint',
 'warning',
 'admonition',
 'sidebar',
 'topic',
 'line-block',
 'parsed-literal',
 'math',
 'rubric',
 'epigraph',
 'highlights',
 'pull-quote',
 'compound',
 'container',
 'table',
 'csv-table',
 'list-table',
 'image',
 'figure',
 'contents',
 'sectnum',
 'header',
 'footer',
 'target-notes',
 'meta',
 'raw',
 'include',
 'replace',
 'unicode',
 'class',
 'role',
 'default-role',
 'title',
 'date',
 'restructuredtext-test-directive']
```

and we know some Sphinx-specific directives (outside docutils) manually that aren't present in this list:

```python
['seealso',
 'versionadded',
 'versionchanged',
 'deprecated',
 'versionremoved',
 'index'
]
```

Only a subset of them will need to be added to the regex pattern, I assume?

agriyakhetarpal: I remembered this `numpydoc` PR from last year that was xref-ed from the NumPy PR where I had reported some incorrect ordering of the Attributes section, as it was showing up after the Examples section. It sheds some light on how the sections are ordered/reordered: https://github.com/numpy/numpydoc/pull/571.

The order preservation is mandated here: https://github.com/numpy/numpydoc/blob/6d5eb428b8acea6ced4d86d086e36daf3f03ffbf/numpydoc/validate.py#L28-L41, and via an error message in https://github.com/numpy/numpydoc/blob/6d5eb428b8acea6ced4d86d086e36daf3f03ffbf/numpydoc/validate.py#L61. We know that dictionaries are ordered in Python post 3.7.

Comparing the relevant code with SymPy's `numpydoc`: https://github.com/sympy/sympy/blob/3c33a8283cd02e2e012b57435bd6212ed915a0bc/doc/ext/docscrape_sphinx.py#L208-L240, not only are the headings and sections different, but also that upstream `numpydoc` has a structural templating mechanism thingy where an order is enforced, and SymPy's version explicitly calls `_str_see_also` at the point where it wants the section to appear. So, if upstream `numpydoc` did not enforce an order, we would also have faced this SymPy-specific problem in NumPy.

Upstream `numpydoc` also has https://github.com/numpy/numpydoc/blob/6d5eb428b8acea6ced4d86d086e36daf3f03ffbf/numpydoc/validate.py#L24-L27 to run checks version-changed directives like the above, but it does not enforce an order for them as they are not a part of standard sections but are directives.

So, I think we should indeed add at least a few of these directives to the patterns we need to ignore – maybe just the most commonly used ones from the above list will work. This would help in both getting SymPy to play nicely, and in the scenario that some docstrings are incorrectly added to any other projects that use the modern `numpydoc`, we don't let that mistake in with the interactive examples.

agriyakhetarpal: At the same time, if someone were to put a `.. caution::` directive in the middle of a doctest, that could cause a bug report for us because the doctest would end midway... but I feel we're technically correct and we could say "works as intended" and close it.

agriyakhetarpal: I've added a minimal set of directives to halt at in commit 79e864e2b5107842fdc2a93d8159f0362245feba and this shouldn't cause any problematic behaviour with most websites, while also getting us working with SymPy.

steppi: I'm still confused. The documentation for `TryExamples` says this

image

so my questions are

1. Can you end up with a raw Sphinx directive inside of the processed docstring when using numpydoc or sphinx.ext.napoleon when using them properly?
2. If the current implementation only works because of the enforced ordering, then it's because it was only working incidentally because of the enforced ordering, because the code is trying to catch all possible next section headers.

Carreau:
Trying to catch up with a lot of tings, but one question I have is: is it worth doing just for sympy ? If there are different way of doing the same things, one of which is the large minority, isn't it better in the long run to update all places that use the least use way ? 

I got that last year with many project using `.. directive[space]::` (while usually there is not space); it was easier to update the few dozen numpy/scipy issues than support the presence of spaces.

Maybe spending the time now to use a newer version of numpydoc would be worth it ?

jupyterlite_sphinx/_try_examples.py
```diff
@@ -303,18 +303,44 @@ def _process_literal_blocks(md_text):
...
+    "contents",
+    "epigraph",
+    "footer",
+    "header",
+    "highlights",
+    "parsed-literal",
+    "pull-quote",
+    "seealso",
+    "sidebar",
+    "topic",
```
steppi, line 326: Some of these like `"pull-quote"`, `"highlights"`, and `"parsed-literal"` seem like they would be reasonable to put in an examples section. Others would be very strange to put in the middle of an examples section when using numpydoc. I made it so that the next section had to be a numpydoc section header because I kept running into new directives that were used in the middle of examples sections. The intention was only to support numpydoc (and sphinx.ext.napoleon because it is compatible) and this is clearly documented. Are there reasonable uses of numpydoc that are not currently supported? I think it's pretty reasonable to say, if you're using `global_enable_try_examples`, and example section is until the next numpydoc section header or the end of the docstring, if you want to delimit the section differently, manually insert the `try_examples` directive.

steppi: OK, here's what I propose. `ignore_directives` should be changed to `explicit_ignore_directives`. If any directive appears in an examples section which is not in `explicit_ignore_directives` or not mapped to markdown in some way, it should be implicity ignored, but there should be an informative warning.

agriyakhetarpal:
Thanks @Carreau and @steppi for the reviews. I think we have two options:

- I can fix this on SymPy's side by porting some rudimentary reordering to SymPy's `numpydoc`. It would be only for this case, so that the See Also section appears before the Examples and we don't change anything else – based on @Carreau's review in https://github.com/jupyterlite/jupyterlite-sphinx/pull/251#issuecomment-2586578533.
- proceed with @steppi's idea of adding `explicit_ignore_directives` and `implicit_ignore_directives` in https://github.com/jupyterlite/jupyterlite-sphinx/pull/251#discussion_r1913096537

The first option has the benefit that we don't need to customise our code to support some other project that we have no control over, and the second option's benefit is that it might lead to fewer issues in the long run and we have the requisite structure for doing it already.

steppi:
> Thanks @Carreau and @steppi for the reviews. I think we have two options:
> 
> A. I can fix this on SymPy's side by porting some rudimentary reordering to SymPy's `numpydoc`. It would be only for this case, so that the See Also section appears before the Examples and we don't change anything else – based on @Carreau's review in [#251 (comment)](https://github.com/jupyterlite/jupyterlite-sphinx/pull/251#issuecomment-2586578533). B. proceed with @steppi's idea of adding `explicit_ignore_directives` and `implicit_ignore_directives` in [#251 (comment)](https://github.com/jupyterlite/jupyterlite-sphinx/pull/251#discussion_r1913096537)
> 
> The first option has the benefit that we don't need to customise our code to support some other project that we have no control over, and the second option's benefit is that it might lead to fewer issues in the long run and we have the requisite structure for doing it already.
> 
> What should we opt for?

I think both A. and B. should be done, but with different priority levels. A is the fix for the immediate issue with SymPy, but there being additional directives which could appear in notebooks shouldn't just be swept aside. It makes more sense to ignore them by default than just keep in the notebook, but in any case, I think this PR could be closed, and B. could be left for later.

agriyakhetarpal:
Albert and I met just some moments ago, and this is our resolution:

- "See Also" heading is also a section that needs to be included – we weren't doing this before, and it is a valid bug as per our current documentation for `TryExamples`. NumPy, SciPy, and co. were saved only because modern `numpydoc` can reorder sections during processing. So, this isn't specific to SymPy at all, but was just caught in an extraordinary case because SymPy did not use modern `numpydoc` and can't reorder the sections.
- More Sphinx directives, such as those for tables and for embedding images/figures can be added, similar to how we convert URLs from reST syntax to Markdown syntax.
- The list of directives to halt/ignore will also be revised, and a warning will be displayed with an invitation to open a new issue if they feel a particular directive is appropriate to use within an "Examples" section, and we can decide ad-hoc on if we want to support that.

Items 2 and 3 are low-priority for now, and I'll open separate issues to track them.

steppi:
> * "See Also" heading is also a section that needs to be included – we weren't doing this before, and it is a valid bug as per our current documentation for `TryExamples`. NumPy, SciPy, and co. were saved only because modern `numpydoc` can reorder sections during processing. So, this isn't specific to SymPy at all, but was just caught in an extraordinary case because SymPy did not use modern `numpydoc` and can't reorder the sections.

I just want to clarify, that the existing code before this PR also considers "See Also" to be a section that should be handled, but it simply has the wrong pattern for matching a "See Also" section. I didn't realize that numpydoc maps a "See Also" section to a `.. seealso:` directive, and just didn't notice this because the order enforcement meant "See Also" couldn't appear after "Examples". The documentation says `TryExamples` should also support `sphinx.ext.napoleon` though, which I think may not have the same order enforcement.

agriyakhetarpal:
Yes, from a quick skim of the code, `sphinx.ext.napoleon` does not enforce a canonical order like `numpydoc` does, and rather, it preserves the order of a particular docstring – because it processes the sections in the order that they appear: https://github.com/sphinx-doc/sphinx/blob/e7dd42ea9110a2eeb8ebc10f4bf8be31f0700b61/sphinx/ext/napoleon/docstring.py#L373-L408

That does offer less consistency and more flexibility in comparison to `numpydoc`, though that's a topic for another discussion.

FWIW, Napoleon also uses the `.. seealso::` directive and creates a section out of it: https://github.com/sphinx-doc/sphinx/blob/e7dd42ea9110a2eeb8ebc10f4bf8be31f0700b61/sphinx/ext/napoleon/docstring.py#L1035-L1036 via https://github.com/sphinx-doc/sphinx/blob/e7dd42ea9110a2eeb8ebc10f4bf8be31f0700b61/sphinx/ext/napoleon/docstring.py#L615-L623, so we should indeed halt at the `.. seealso::` directive that has now been done in this PR; this change covers both `sphinx.ext.napoleon` and `numpydoc`.

steppi (approved):
Thanks @agriyakhetarpal. This looks good to me now. I can work on the follow up for handling other section headers which aren't mapped to `admonition` or `rubric`, and for ignoring other unhandled directives. I feel a sense of responsibility over these things since I was the one who missed them when implementing `try_examples`

PR #267, opened 2025-02-04
Update to `actions/upload-artifact@v4`

jtpio:
This should help fix the failing CI.

### DIFF
--- a/.github/workflows/check-release.yml
+++ b/.github/workflows/check-release.yml
@@ -18,12 +18,9 @@ jobs:
     timeout-minutes: 30
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Base Setup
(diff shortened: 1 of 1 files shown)


.github/workflows/check-release.yml
```diff
@@ -18,20 +18,17 @@ jobs:
     timeout-minutes: 30
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Base Setup
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
-      - name: Install Dependencies
```
jtpio, line 24: Also taking the opportunity to remove this step as it should not be needed. It can be even be leading to the wrong assumption the repo is in a "releasable" state, since the real release workflow don't perform this step.

Carreau (approved):
Thanks for taking the time to do this.

PR #277, opened 2025-03-24
pin micromamba to v2.0.5 to fix failing RTD builds

### LINKED ISSUE #276: "latest" builds are failing on RTD
https://app.readthedocs.org/projects/jupyterlite-sphinx/builds/27616246/

Also noticed in #274.

agriyakhetarpal:
This PR fixes #276. The pin should be temporary.

### DIFF
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -8,7 +8,7 @@ dependencies:
     - jupyterlite-core >=0.3,<0.6
     - jupytext
     - pydata-sphinx-theme
-    - micromamba
+    - micromamba=2.0.5
     - myst-parser
     - docutils
     - sphinx
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,7 +18,7 @@ strip_tagged_cells = True
 
 # Enable this to unsilence JupyterLite and aid debugging
 # within our own documentation.
-# jupyterlite_silence = False
+jupyterlite_silence = False
 
 master_doc = "index"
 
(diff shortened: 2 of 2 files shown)


agriyakhetarpal:
The prefix is set to `/home/docs/checkouts/readthedocs.org/user_builds/jupyterlite-sphinx/conda/jupyterlite-sphinx` instead of `/tmp/tmp5a5b_ed1/env/envs/jupyterlite-sphinx`, which means that the kernels can't be copied.

This does not occur locally, as the prefix shows up as something like `/private/var/folders/b3/2bq1m1_50bs4c7305j8vxcqr0000gn/T/tmpgxb2987u/env/envs/jupyterlite-sphinx`.

Read the Docs is overriding the prefix somewhere in its build process, which is also the cause for several `does not contain any filesystem separator` warnings. `jupyterlite-xeus` looks for the prefix elsewhere, and can't find this overridden one. I guess the fix should be for `jupyterlite-xeus` to respect the standard `conda`/`mamba` environment variables that handle prefixes.

martinRenou:
Thanks for debugging this! Can you try pinning to micromamba 2.0.5 in the dev env? https://github.com/jupyterlite/xeus/pull/199

I've seen issues with latest micromamba, although I don't have time to debug this so pinning is the way I do for now.

Note that the coming jupyterlite-xeus 4.0.0 release will depend on micromamba directly, and pin it properly.

agriyakhetarpal:
Thanks, @martinRenou – that works for now!

PR #280, opened 2025-04-11
Allow adding contents from outside the Sphinx source directory

### LINKED ISSUE #261: jupyterlite_contents paths not picked up if run from outside of app.srcdir
When the top build command is run outside of Sphinx' `srcdir` (which is the case when buliding from Jupyterbook for instance) the following happens:

1. [`jupyterlite_contents` path are expanded with glob using the cwd as root](https://github.com/jupyterlite/jupyterlite-sphinx/blob/42a9b1c479cc89f60057d0292c42f5d63445caa6/jupyterlite_sphinx/jupyterlite_sphinx.py#L992). This means that paths that don't exist after expansion are removed from the list.
2. [`jupyter lite build` is invoked](https://github.com/jupyterlite/jupyterlite-sphinx/blob/42a9b1c479cc89f60057d0292c42f5d63445caa6/jupyterlite_sphinx/jupyterlite_sphinx.py#L1050) using `app.srcdir` as cwd.

So say I have the following structure

```text
repo
└── book
    ├── conf.py
    └── notebooks
        └── nb.ipynb
```

To build the book, I have to run Jupyterbook in `repo` and I want `notebooks/` to be in `jupyterlite_contents`.

- If I set `jupyterlite_contents = ["book/notebooks"]`, `glob` in 1. will be able to find it since for glob, the root dir is `repo`, but when `jupyter lite build` runs in 2., since its cwd is `book`, it won't find `book/notebooks` and the build will fail.
- If I set `jupyterlite_contents = ["notebooks"]` instead, `glob` in 1. won't find it, sot it will silently remove it from the list. The build will succeed but the `notebook` won't be picked up.

Either way, I can't get `notebook` in my build if it's not run from `app.srcdir`.

A simple fix would be to set `glob`'s `rootdir` parameter to `src.appdir`. It would still allow adding paths outside of the appdir using `..`. I can send a PR as soon as I have you go, or as soon as we figure out another solution if this one doesn't work, as I am **very** interested in having this fixed :-)

agriyakhetarpal:
### Description

This PR allows the `jupyterlite_contents` configuration option to pick up contents from `app.srcdir`, i.e., to add notebooks outside `docs/` or `docs/source/`. For example, it can be used to include a directory from outside the docs, say, in an `examples/` folder that's in the repository's root folder (a pattern that's often the convention in the structure of a Python package) or any other path on their machine.

Support for including paths inside the srcdir and for glob patterns is retained.

Therefore, users may now specify the following in `conf.py`

```py
jupyterlite_contents = [
  "./custom_contents/",
  "../my_example_dir/**",
  "../my_other_example_dir/**",
  "/Users/username/absolute/path/to/notebook.ipynb",
]
```

and so on, or absolute paths on their machine if they wish to.

<hr>

Closes #261
This also came up in statsmodels/statsmodels#9536

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
     types_or: [css, javascript]
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.11.4
+  rev: v0.11.7
   hooks:
     # Run the linter.
     - id: ruff
--- a/docs/directives/try_examples.md
+++ b/docs/directives/try_examples.md
@@ -195,7 +195,7 @@ if needed. Adding the comment
 
 
 ```rst
-..! disable_try_examples`
+.. disable_try_examples
 ```
 
 as the first non-empty line under
--- a/jupyterlite_sphinx/_try_examples.py
+++ b/jupyterlite_sphinx/_try_examples.py
@@ -339,7 +339,7 @@ def insert_try_examples_directive(lines, **options):
     """Adds try_examples directive to Examples section of a docstring.
 
     Hack to allow for a config option to enable try_examples functionality
-    in all Examples sections (unless a comment "..! disable_try_examples" is
+    in all Examples sections (unless a comment ".. disable_try_examples" is
     added explicitly after the section header.)
 
 
@@ -354,7 +354,7 @@ def insert_try_examples_directive(lines, **options):
     list of str
         Updated version of the input docstring which has a try_examples directive
         inserted in the Examples section (if one exists) with all Examples content
-        indented beneath it. Does nothing if the comment "..! disable_try_examples"
+        indented beneath it. Does nothing if the comment ".. disable_try_examples"
         is included at the top of the Examples section. Also a no-op if the
         try_examples directive is already included.
     """
@@ -375,8 +375,8 @@ def insert_try_examples_directive(lines, **options):
         # Examples section had no content, no need to insert directive.
         return lines[:]
 
-    # Check for the "..! disable_try_examples" comment.
-    if lines[left_index].strip() == "..! disable_try_examples::":
+    # Check for the ".. disable_try_examples" comment.
+    if lines[left_index].strip() == ".. disable_try_examples":
         # If so, do not insert directive.
         return lines[:]
 


jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -990,15 +989,38 @@ def jupyterlite_build(app: Sphinx, error):
...
+            pattern_path = Path(pattern)
+
+            if pattern_path.is_absolute():
+                base_path = pattern_path.parent
+                glob_pattern = pattern_path.name
+            else:
+                # For relative paths, we'll resolve them relative
+                # to app.srcdir
+                base_path = Path(app.srcdir) / pattern_path.parent
+                glob_pattern = pattern_path.name
```
steppi, line 1003:
```suggestion
            base_path = pattern_path.parent if pattern_path.is_abolute() else Path(app.srcdir) / pattern_path.parent
            glob_pattern = pattern_path.name
```

`glob_pattern` is being set to the same thing regardless of which branch is taken, so it can be moved out of the `if-else` block.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -990,15 +989,38 @@ def jupyterlite_build(app: Sphinx, error):
...
+
+            if any(c in str(glob_pattern) for c in "*?[]"):
+                matched_paths = list(base_path.glob(glob_pattern))
+            else:
+                full_path = base_path / glob_pattern
+                matched_paths = [full_path] if full_path.exists() else []
```
steppi, line 1009:
```suggestion
            matched_paths = base_path.glob(glob_pattern)
```

I'm not sure its worth adding all of these extra lines just to avoid calling `glob` when there are no special characters in the pattern. I can't imagine it makes any appreciable difference in runtime. Also, you don't need to convert it into a list because you're just going to iterate through it later anyway.

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -990,15 +989,38 @@ def jupyterlite_build(app: Sphinx, error):
...
+                # We need to provide paths that will work when the command runs
+                # with cwd=app.srcdir. So, we compute the relative path from
+                # app.srcdir to each match
+                if match_path.is_absolute():
+                    rel_path = str(match_path)
+                else:
+                    try:
+                        rel_path = str(match_path.relative_to(Path(app.srcdir)))
+                    except ValueError:
+                        rel_path = str(match_path.absolute())
```
steppi, line 1021: I'm don't understand what's happening here. Can you walk me through it? If `match_path` is absolute, how can it be the correct relative path from `app.srcdir`, or are the above comment and variable name `rel_path` both out of date? Also, what situations do you expect to lead to a `ValueError` error here, and is the right thing to do in that case really to convert `match_path` to an absolute path.

steppi (changes requested):
I had a couple minor comments about simplifying the conditional logic.

There's one block of code that I'm having trouble understanding, possibly because the comment is out of date.

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: I think the variable name is confusing; I should have used something like `final_path` or `contents_path` instead of `rel_path`. I'll switch.

The only scenario where the `except ValueError:` branch would be necessary is if _somehow_ `match_path` is both not absolute AND not within the Sphinx source directory tree, which seems unlikely to happen given how we construct the paths – my oversight! I don't think we need to branch out for that.

At the moment, the only use case is to allow adding contents from outside `app.srcdir`, whether absolute or relative, and the vast majority users would use relative paths if inside `app.srcdir` anyway.

steppi (approved):
Thanks @agriyakhetarpal. This looks good to me now.

PR #292, opened 2025-05-08
Fix `IndexError`s when the "Examples" is the last section in a docstring

### LINKED ISSUE #291: `try_examples` fails if the examples sections is last
## Description

When building the sphinx documentation with `sphinx.ext.napoleon` and if `Examples` is the last section, `try_examples` appears exceed the number of lines in the docstring:
https://github.com/jupyterlite/jupyterlite-sphinx/blob/d2a34cb0bc8a8d95263b4b1329a82f801440021f/jupyterlite_sphinx/_try_examples.py#L390-L400

If there's no section pattern before the end of the docstring, `right_index` will be `len(lines)` in line 394, causing the `__getitem__` call to fail with an `IndexError`. Since the check in line 394 appears to be checking for a special section pattern, it should be fine to just skip the entire block by prepending `right_index < len(lines)` to the condition. However, that means that `new_lines` will contain a artificial trailing empty lines, so that would most likely have to be trimmed.

## Reproduce

`conf.py`:
```python
# Configuration file for the Sphinx documentation builder.
#
# For the full list of built-in configuration values, see the documentation:
# https://www.sphinx-doc.org/en/master/usage/configuration.html

# -- Project information -----------------------------------------------------
# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information

project = 'test'
copyright = '2025, testing'
author = 'testing'

import os
import sys

sys.path.insert(0, os.path.abspath("."))

# -- General configuration ---------------------------------------------------
# https://www.sphinx-doc.org/en/master/usage/configuration.html#general-configuration

extensions = ["jupyterlite_sphinx", "sphinx.ext.autosummary", "sphinx.ext.napoleon"]

templates_path = ['_templates']
exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']

global_enable_try_examples = True



# -- Options for HTML output -------------------------------------------------
# https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-output

html_theme = 'alabaster'
html_static_path = ['_static']
```

`example.py`
```python
#!/usr/bin/env python3

def test(a):
    """
    Parameters
    ----------
    a : int
        first value

    Examples
    --------
    >>> test(1)
    3
    """

    return a * 2 + 1
```

`index.rst`:
```rst
Welcome to test's documentation!
================================

.. toctree::
   :maxdepth: 2
   :caption: Contents:

.. autosummary::
   :toctree: generated/

   example.test
```
## Expected behavior

No error.

## Context

ubuntu                   24.04
python                    3.12.10         h9e4cc4f_0_cpython    conda-forge
jupyterlite-sphinx        0.20.0             pyhd8ed1ab_0    conda-forge

agriyakhetarpal:
This PR closes #291, it fixes an `IndexError` that occurs when processing docstrings where the Examples section is the last section in a docstring when using `sphinx.ext.napoleon`. I've added bounds checking to prevent accessing indices beyond the end of the lines list in the `insert_try_examples_directive` function.

### DIFF
--- a/jupyterlite_sphinx/_try_examples.py
+++ b/jupyterlite_sphinx/_try_examples.py
@@ -408,11 +408,10 @@ def insert_try_examples_directive(lines, **options):
         + [f"    :{key}: {value}" for key, value in options.items()]
         + [""]
         + ["    " + line for line in lines[left_index:right_index]]
-        + [""]
     )
 
     # Append the remainder of the docstring, if there is any
     if right_index < len(lines):
-        new_lines += lines[right_index:]
+        new_lines += [""] + lines[right_index:]
 
     return new_lines
--- a/jupyterlite_sphinx/_try_examples.py
+++ b/jupyterlite_sphinx/_try_examples.py
@@ -391,12 +391,14 @@ def insert_try_examples_directive(lines, **options):
         lines[right_index]
     ):
         right_index += 1
-    if "!! processed by numpydoc !!" in lines[right_index]:
+
+    # Check if we've reached the end of the docstring
+    if right_index < len(lines) and "!! processed by numpydoc !!" in lines[right_index]:
         # Sometimes the .. appears on an earlier line than !! processed by numpydoc !!
         if not re.search(
             r"\.\.\s+\!\! processed by numpy doc \!\!", lines[right_index]
         ):
-            while lines[right_index].strip() != "..":
+            while right_index > 0 and lines[right_index].strip() != "..":
                 right_index -= 1
 
     # Add the ".. try_examples::" directive and indent the content of the Examples section
@@ -407,7 +409,10 @@ def insert_try_examples_directive(lines, **options):
         + [""]
         + ["    " + line for line in lines[left_index:right_index]]
         + [""]
-        + lines[right_index:]
     )
+
+    # Append the remainder of the docstring, if there is any
+    if right_index < len(lines):
+        new_lines += lines[right_index:]
 
     return new_lines


jupyterlite_sphinx/_try_examples.py
```diff
@@ -407,7 +409,10 @@ def insert_try_examples_directive(lines, **options):
         + [""]
         + ["    " + line for line in lines[left_index:right_index]]
         + [""]
-        + lines[right_index:]
     )
 
+    # Append the remainder of the docstring, if there is any
+    if right_index < len(lines):
+        new_lines += lines[right_index:]
```
keewis, line 416: should this also include the empty line you're appending, i.e.:
```suggestion
    # Append the remainder of the docstring, if there is any
    if right_index < len(lines):
        new_lines += [""] + lines[right_index:]
```
(and remove the one above, at the end of the construction of `new_lines`)

agriyakhetarpal: Ah, makes sense to me!

agriyakhetarpal:
The errors look unrelated, and I think GitHub has degraded performance at the moment – we should be fine.

agriyakhetarpal:
Okay, the error is apparently on pre-commit.ci's side; we should be good to go here as the RTD build is working.

jupyterlite_sphinx/_try_examples.py
```diff
@@ -406,8 +408,10 @@ def insert_try_examples_directive(lines, **options):
         + [f"    :{key}: {value}" for key, value in options.items()]
         + [""]
         + ["    " + line for line in lines[left_index:right_index]]
-        + [""]
-        + lines[right_index:]
     )
 
+    # Append the remainder of the docstring, if there is any
+    if right_index < len(lines):
```
steppi, line 414: To make sure I understand correctly, this condition will be False in the case where the example section is last and there is no emptyline after it. The check keeps `jupyterlite_sphinx` from inserting an emptyline at the end in this case. If so, that sounds reasonable.

steppi (approved):
Looks good @agriyakhetarpal, thanks for fixing this.

agriyakhetarpal:
I still can't figure out what the issue with the check release job is, but it's not related, obviously, so we can debug it as a part of the release process.

agriyakhetarpal:
Thanks for the review!

PR #290, opened 2025-05-07
Add more REPL customisation features for JupyterLite 0.6.0 and later

agriyakhetarpal:
## Description

This PR closes #275; it adds the new REPL configuration options for the Replite directive from https://github.com/jupyterlite/jupyterlite/pull/1573 and associated documentation for these options. Like the previous addition via #245, these options are also configured via URL parameters and can be configured either globally or via a per-directive basis. Many of these options are in camelCase, and are converted to snake_case accordingly.

PR #299, opened 2025-06-03
Bump jupyterlite-core dependency

martinRenou:


jtpio:
A patch (`0.20.2`) sounds good since this change is about loosening the Python dependency.

There is https://github.com/jupyterlite/jupyterlite-sphinx/pull/290 related to the `0.6.0` update, but it can go in `0.21.0`.

martinRenou:
That's also what I had in mind 👍🏽 I'll go ahead and do it

PR #309, opened 2025-08-27
Add `jupyterlite_ignore_contents` config, mapping to `--ignore-contents` arg in JupyterLite build

mfisher87:
Resolves #308

PR #311, opened 2025-08-27
Run `black` at pre-commit time

mfisher87:
My PR #309 passed pre-commit checks, but failed when GitHub Actions ran `black --check`. Running `black` at pre-commit time is a bit less frictiony :)

I considered using Ruff for formatting, but Ruff and Black have, I believe, diverged slightly. I think switching CI + pre-commit to both use Ruff for formatting would be better done in a separate PR.

### DIFF
--- a/.github/workflows/check-release.yml
+++ b/.github/workflows/check-release.yml
@@ -5,8 +5,7 @@ on:
   pull_request:
 
 permissions:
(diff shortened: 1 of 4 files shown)


martinRenou (approved):
Thanks!

Issue #314, opened 2025-09-17
Bypass "select kernel" dialog


chrysn:
### Problem

The first interaction I have with a `replite` instance is always to select a kernel -- even if only one is installed, or when the page author already made a selection. This stops the reading and exercising flow.

### Proposed Solution

* If there is just one kernel available, use that.
* Allow the page author to specify the kernel.
  [edit, added:] There *is* a kernel configuration parameter as demo'd at <https://jupyterlite-sphinx.readthedocs.io/en/stable/directives/replite.html>, but it has no apparent effect there.

### Additional context

I'm looking into using [jupyterlite-sphinx for aiocoap's documentation](https://codeberg.org/aiocoap/aiocoap/pulls/23), which [currently](https://aiocoap.readthedocs.io/en/latest/guidedtour.html#a-request) has a lot of Python examples just in plain text, but I'd love to show off how the library works [well with Jupyter](https://aiocoap.readthedocs.io/en/latest/pyodide.html) already live to the user, possibly even making the guided tour fully exercisable in the browser.

chrysn:
It is supported in JupyterLite as a query parameter:

* No prompt: https://jupyterlite-pyodide-kernel.readthedocs.io/en/latest/_static/repl/index.html?kernel=python
* Same prompt: https://jupyterlite-pyodide-kernel.readthedocs.io/en/latest/_static/repl/index.html

For the concrete example of https://jupyterlite-sphinx.readthedocs.io/en/stable/directives/replite.html, maybe the documentation just has a wrong kernel value: Placing xpython rather than xeus-python in the argument seems to already do the right thing.

(Something that might be relevant to add upstream is to have some kernel=auto parameter, or if that's default, kernel=ask to explicitly show the popup, but probably this can be fixed in documentation by giving concrete values and examples for kernel.)

PR #316, opened 2025-09-19
doc: Explain repl's kernel parameter, fix value to avoid dialog

chrysn:
Closes: https://github.com/jupyterlite/jupyterlite-sphinx/issues/314

The current REPL examples all needlessly have the "select a kernel" dialog pop up, because the name given doesn't match the installed kernel. This PR fixes the names, and adds an explanation that mentions the dialog


martinRenou (approved):
Thanks!

PR #317, opened 2025-09-19
replite: Support `:showBanner: 0`

chrysn:
This adds support for hiding the REPL kernel's banner, which takes up quite a bit of screen real estate for xeus, and at least a line for pyodide. Hiding the banner is particularly desirable for documentation with many short REPL snippets.


martinRenou (approved):
Neat! Thanks

PR #333, opened 2026-02-24
Fix overflow for JupyterLite iframes when added by directives used inside PST admonitions

agriyakhetarpal:
This PR fixes the visual bug I identified in point 1 of https://github.com/jupyterlite/jupyterlite-sphinx/pull/290#issuecomment-3955068843 about the JupyterLite REPL (and other apps') iframes overflowing if they are part of admonitions. This is a hack right now and may or may not work with other themes, but we've never marketed official support of the extension for themes beyond the PST anyway.

Since #290 is not merged yet, I applied these changes on top of it locally, and the following screenshots showcase the difference/fix (please ignore the non-working iframe in the "After" image, that's not related):

| Before ( | After |
|--------|--------|
| image | image | 

<hr>

also cc: @Carreau for visibility; as this involves the PST


martinRenou (approved):
Thanks!

PR #334, opened 2026-02-24
Drop support for Python 3.9

agriyakhetarpal:
As @keewis pointed out to me in #332, Python 3.9 is now EOL. It reached this status on 31 October, 2025, as mentioned on https://endoflife.date/python.


mfisher87 (approved):
LGTM :) Let's also update `.github/workflows/main.yml` before merge.

Issue #329, opened 2026-01-27
Make jupytext an optional dependency


martinRenou:
### Problem

I'd like to suggest that we make jupytext an optional dependency. 

The problem I see is that jupytext is incompatible with JupyterLite today, and pulling it in the environment pulls the labextension for it, that JupyterLite will pick up automatically, even though it's only needed at build time here (for processing markdown files).

@agriyakhetarpal what do you think?

PR #335, opened 2026-02-25
Use dependency groups instead of optional dependencies

agriyakhetarpal:
These were never meant to be listed in public metadata anyway, so we can avoid doing that now :D This is also in prep for #329.

No other changes should be needed – I checked `.readthedocs.yaml` (https://docs.readthedocs.com/platform/latest/build-customization.html#install-dependencies-from-dependency-groups) but this doesn't apply to us as we list all our docs build dependencies directly in the requirements there. I don't see any other places where we mention `[docs]`, `[dev]`, `[dev,docs]`, etc. either.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,11 +22,10 @@ dependencies = [
     "sphinx>=4",
 ]
 
-[project.optional-dependencies]
+[dependency-groups]
 dev = [
     "hatch",
 ]
-
 docs = [
     "myst_parser",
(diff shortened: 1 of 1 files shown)


mfisher87:
Yessss :) :100: :100: :100:

mfisher87 (approved):

agriyakhetarpal:
Awesome, thanks for the super quick review, @mfisher87!

PR #336, opened 2026-02-25
Make `jupytext` an optional dependency

### LINKED ISSUE #258: Better separation of `jupyterlite-sphinx`'s dependencies and that of the JupyterLite environment
Ah, I understand; yes, it would be nice not to inject `jupytext` into the JupyterLite environment and not dilute any extensions. I did suggest keeping it as an optional dependency in https://github.com/jupyterlite/jupyterlite-sphinx/issues/191#issuecomment-224458997 under a flag, but no one talked about it in the code review, so I suppose it was not addressed.

It makes me think that we are constructing JupyterLite instances with shared dependencies (such as from a project's `[docs]` extras), which also creates the possibility of dependency conflicts that are not exposed to the user, besides such corruptions in the JupyterLite environment that you shared. As a solution, how would you feel about "separating" the Sphinx docs builds and the JupyterLite builds better – say, through constructing a virtual environment with `virtualenv` and activating it, installing JupyterLite and any extensions into that environment separately, and then discarding said environment at the end of the Sphinx build when we have all the files? This way, `jupytext` would be retained for converting notebooks from Markdown to IPyNB, but wouldn't have a role with JupyterLite, as it would not exist when the static files are built – further limiting its usage.

We could then have configuration options to control the version of JupyterLite, and other dependencies (such as `juptyerlite-pyodide-kernel`, etc.) including their versions from `conf.py` itself:

```python
jupyterlite_version = "0.5.0"

# or maybe construct a conda env and add "jupyterlite_conda_dependencies" as well?
jupyterlite_pypi_dependencies = [
"jupyterlab_materialdarker>=0.6.0",
"jupytext==1.14.0"
]

# maybe? or is the recommended way to install nowadays just "pip"?
jupyterlite_npm_dependencies = [
"base16-gruvbox-dark"
]
```

It might look like a bit much, but could be more correct in the long run. Still, it might be possible to find a middle ground in terms of the complexity.

_Originally posted by @agriyakhetarpal in https://github.com/jupyterlite/jupyterlite-sphinx/issues/236#issuecomment-2561926161_

### LINKED ISSUE #329: Make jupytext an optional dependency

agriyakhetarpal:
Closes #329
Closes #258

### DIFF
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -28,8 +28,12 @@ from ._try_examples import (
     new_code_cell,
 )
 
-import jupytext
 import nbformat
+
+try:
+    import jupytext
+except ImportError:
+    jupytext = None
 
 try:
     import voici
@@ -617,6 +621,11 @@ class _LiteDirective(SphinxDirective):
             notebook_is_stripped: bool = self.env.config.strip_tagged_cells
 
             if notebook_path.suffix.lower() == ".md":
+                if jupytext is None:
+                    raise ImportError(
+                        "jupyterlite-sphinx requires the jupytext package to process Markdown notebooks. "
+                        'Install it with: pip install "jupyterlite-sphinx[markdown]"'
+                    )
                 if self._target_is_stale(notebook_path, target_path):
                     nb = jupytext.read(str(notebook_path))
                     if notebook_is_stripped:
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -624,7 +624,7 @@ class _LiteDirective(SphinxDirective):
                 if jupytext is None:
                     raise ImportError(
                         "jupyterlite-sphinx requires the jupytext package to process Markdown notebooks. "
-                        'Install it with: pip install "jupyterlite-sphinx[markdown]"'
+                        'Install "jupyterlite-sphinx[markdown]" with your package manager of choice.'
                     )
                 if self._target_is_stale(notebook_path, target_path):
                     nb = jupytext.read(str(notebook_path))
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -15,5 +15,5 @@ dependencies:
   - black
   - voici
   - pip:
-      - .
+      - ."[markdown]"
       - jupyterlite-xeus >=2.1.2
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -15,5 +15,5 @@ dependencies:
   - black
   - voici
   - pip:
-      - ."[markdown]"
+      - .[markdown]
       - jupyterlite-xeus >=2.1.2
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,12 +17,14 @@ dependencies = [
     "jupyter_server",
     "jupyterlab_server",
     "jupyterlite-core >=0.2,<0.8",
-    "jupytext",
     "nbformat",
     "sphinx>=4",
 ]
 
 [project.optional-dependencies]
+markdown = ["jupytext"]
+
+[dependency-groups]
 dev = [
     "hatch",
 ]


agriyakhetarpal:
Hi @mfisher87, I requested a review from you too, assuming that you are interested in reviewing PRs going forward! Thanks for your work on `jupyterlite-sphinx` towards the latter half of last year when I was (sort of) absent from development for a while 😄

jupyterlite_sphinx/jupyterlite_sphinx.py
```diff
@@ -617,6 +621,11 @@ def run(self):
             notebook_is_stripped: bool = self.env.config.strip_tagged_cells
 
             if notebook_path.suffix.lower() == ".md":
+                if jupytext is None:
+                    raise ImportError(
+                        "jupyterlite-sphinx requires the jupytext package to process Markdown notebooks. "
+                        'Install it with: pip install "jupyterlite-sphinx[markdown]"'
```
mfisher87:
```suggestion
                        'Install "jupyterlite-sphinx[markdown]" with your package manager of choice.'
```

With the rate of `uv` adoption I wonder if we should avoid pip-specific instructions? Either way is fine by me! :)

mfisher87 (approved):
> Hi @mfisher87, I requested a review from you too, assuming that you are interested in reviewing PRs going forward! Thanks for your work on `jupyterlite-sphinx` towards the latter half of last year when I was (sort of) absent from development for a while 😄

Happy to review! LGTM! One bit of feedback.

jupyterlite_sphinx/jupyterlite_sphinx.py
agriyakhetarpal: Makes sense for the instructions to be agnostic to package managers! Your suggestion also helps with conda-forge. Thanks!

mfisher87: Although unfortunately conda doesn't support an extras syntax :(

agriyakhetarpal: Ah, right. I think we will have to keep it as a dependency there (I don't think it's worth the effort to set up `jupyterlite-sphinx-base` and then add another build output or any of those shenanigans). In that case, conda-forge users will get the dependency and never reach this scenario.

Perhaps this will be better?

> "Install `jupyterlite-sphinx[markdown]` or `jupytext` with your package manager of choice."

mfisher87: If we're keeping jupytext as a dependency of the conda-forge package, I think we're good with the message as-is (conda-forge users won't see it), but this is also OK :)

agriyakhetarpal: On second thought, we will have to think through this a bit, actually. @jtpio previously had some thoughts about Jupytext being included in https://github.com/jupyterlite/jupyterlite-sphinx/pull/236#issuecomment-2590610993, and I believe that it won't be helpful to keep it as a dependency on `conda-forge` either, as users from there will still see (and have been seeing) Jupytext popping up in the JupyterLite launcher at a time when Jupytext is not yet supported in JupyterLite (https://github.com/mwouts/jupytext/issues/1225). We should evaluate whether we should remove the dependency from conda-forge as well. One user of `jupyterlite-sphinx`'s converted-Markdown-notebooks functionality from `conda-forge` is SciPy, but there may be others as well.

martinRenou (approved):
🚀

jupyterlite_sphinx/jupyterlite_sphinx.py
martinRenou: I'm 👎🏽 on having jupytext as a dependency in the conda-forge package. Unfortunately conda-forge does not have a notion of extras, so we can provide some `jupyterlite-sphinx-markdown` package that is a metapackage pulling `jupyterlite-sphinx` and `jupytext`, making the feedstock of jupyterlite-sphinx a multi-ouput recipe.

agriyakhetarpal: Yes, let's do that! I don't think I've actually worked with multi-output recipes and metapackages in conda-forge before, but I think it shouldn't be too hard.

martinRenou: I can provide help. You will need to make an admin request to conda-forge to add a new output to the feedstock, see https://github.com/conda-forge/admin-requests/pull/1567 for example

agriyakhetarpal: Gotcha, thanks! I'll open an issue in the jupyterlite-sphinx feedstock to remind ourselves of this when we do the release.

Issue #307, opened 2025-08-26
I want to specify a nested directory structure as contents


mfisher87:
### Problem

Story: I want to expose a directory structure in my jupyterlite instance. This jupyterlite instance is a part of a documentation site, and I want the jupyterlite instance to help users learn about the software I'm documenting. I want the top level of the jupyterlite directory structure to contain notebooks introducing concepts in a specific order. And I want to separate the data those notebooks use into a `data/` subdirectory. E.g.:

```
.
├── 01-summary.ipynb
├── 02-basic-api.ipynb
├── 03-advanced-api.ipynb
└── data
    ├── bar
    ├── baz
    │   ├── baz1
    │   └── baz2
    └── foo
```

When I set:

```
jupyterlite_contents = [
    "../examples/*",
]
```

My JupyterLite instance has a `files` directory structure like:

```
files/
├── 01-summary.ipynb
├── 02-basic-api.ipynb
├── 03-advanced-api.ipynb
├── bar
├── baz
 |   ├── baz1
 |   └── baz2
└── foo
```

(The contents of the "data" directory are flattened out to the root of `files/`)

When I set:

```
jupyterlite_contents = [
    "../examples/**",
]
```

I get in my JupyterLite deployment:

```
files/
├── 01-summary.ipynb
├── 02-basic-api.ipynb
├── 03-advanced-api.ipynb
├── bar
├── baz
 |   ├── baz1
 |   └── baz2
├── baz1
├── baz2
└── data
 |   ├── bar
 |   ├── baz
 |   │   ├── baz1
 |   │   └── baz2
 |   └── foo
└── foo
```

(Every file and directory under `data/` is recursively duplicated to their parent directory, so e.g. `baz1` shows up 3 times)

I tried a few other permutations, and nothing gave the desired result of an exact recursive copy of the target directory.

### Proposed Solution

* Iterate over the glob expression

mfisher87:
Well... one thing I didn't try was a directory _without_ a glob expression:

```
jupyterlite_contents = [
    "../examples",
]
```

This does what I want! However, the behavior of duplicating files when using a glob is still surprising.

PR #337, opened 2026-02-25
Un-flatten JupyterLite contents: preserve notebook paths and directory structure

agriyakhetarpal:
Previously, when support for Markdown-based notebooks was added in #221, we added mechanisms to handle potential notebook name collisions and hoped that no one in the wild would have two notebooks of the same name across their docs directories and subdirectories. This is because we currently "flatten" all notebooks marked for inclusion on the JupyterLite site via directives into a single `/path/to/docs/lite/files/` folder. I have started to think that it is a bad idea, since the contemporary Sphinx-Gallery does not do that as part of its JupyterLite integration and preserves the entire nested structure of the notebooks.

This PR stops doing so. We can now have notebooks with the same filename, say, `my_notebook.ipynb`, placed in two distinct folders, as the directory structure will be preserved. We can't have notebooks of the same filename under the same folder anyway, as that will be restricted by the file system :D

I have removed the `_assert_no_conflicting_nb_names` check and preserved relative notebook paths when converting/copying. Additionally, `jupyterlite_contents` entries that are directories are now copied into the `_contents/` staging area rather than being passed directly as `--contents` arguments to `jupyter lite build`. This will help preserve the directory name in the JupyterLite file system: previously, passing `--contents custom_contents` caused JupyterLite to treat that directory as a content root, placing its files at the root of files/ rather than under files/custom_contents/. Routing them through `_contents/`, which is already passed as a single `--contents` argument, ensures the directory name will be retained.


agriyakhetarpal:
@mfisher87, I noticed that you opened https://github.com/jupyterlite/jupyterlite-sphinx/issues/307. This is _sort of_ related to the issue, since it should partially fix the flattening out of the `examples/` directory you had. I would appreciate you taking a look, so that we don't have any unexpected behaviour and can catch things in advance for JupyterGIS! 😄 That said, you mentioned including some info in the docs about how the globs work when specified in `jupyterlite_contents`. I would welcome such a PR post these changes, however they land in the end. I myself have had some confusion on how `jupyterlite_contents` should work on a few occasions in the past.

- I believe the case where `examples/**` is specified duplicates files will remain with this PR
- If `../examples` (without any glob) is specified, the files will appear under `files/examples/` and won't be flattened, but I am not sure if any users are relying on the current behaviour in weird ways

martinRenou:
I took the liberty to rebase and fix conflicts, hope you don't mind!

PR #339, opened 2026-02-27
Unpin `jupyterlite-xeus` in docs dependency group and use latest micromamba for docs builds

agriyakhetarpal:
Trying out https://github.com/jupyterlite/xeus/pull/199 here to see if we can unpin these dependencies now. Note that jupyterlite-xeus was not pinned in RTD as we don't use dependency groups, the pin just affected developers like me who use `uv`/`pip` to manage their environments. micromamba was pinned in #277 has been reverted now.

### DIFF
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -8,7 +8,7 @@ dependencies:
   - jupyterlite-core >=0.3,<0.7
   - jupytext
   - pydata-sphinx-theme
-  - micromamba=2.0.5
+  - micromamba
   - myst-parser
   - docutils
   - sphinx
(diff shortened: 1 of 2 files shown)


agriyakhetarpal:
Looks like we are all green (minus the check-release workflow that I already tracked in https://github.com/jupyter-server/jupyter_releaser/issues/645), thanks @martinRenou, assuming you were the one to fix this in micromamba!

martinRenou:
I did not fix anything in micromamba, I used the ancestral technique to wait a year for the bug to disappear by itself 🙉 

That reminds me that I should make a jupyterlite-xeus release and unpin micromamba in the conda package, on it now. Thanks!

PR #346, opened 2026-07-28
allow configuring the path of the contents directory

keewis:
This is the one remaining issue I have before being able to merge the PR that adds `try_examples` to the `xarray` docs.

Basically, I would run sphinx once and then retry again after making changes, and if the first run exited with a non-zero exit code the second run would try to execute all the notebooks in the `_contents` directory. However, if I were to add `_contents` to `exclude_patterns` `jupyterlite` wouldn't be able to find the notebooks in `_contents`.

As far as I can tell, changing the path from `_contents` to `_build/contents` (or `_build/jupyterlite-contents`) would always succeed, and wouldn't get confused by `exclude_patterns`. So what I did was:
1. make the path configurable
2. clean up _before the next run_ instead of _after the current run_

cc @agriyakhetarpal

### DIFF
--- a/jupyterlite_sphinx/jupyterlite_sphinx.py
+++ b/jupyterlite_sphinx/jupyterlite_sphinx.py
@@ -961,8 +961,13 @@ def conditional_process_examples(app, config):
 
 
 def inited(app: Sphinx, config):
+    if not app.config.jupyterlite_content_dir:
+        raise ValueError("jupyterlite_content_dir must be a non-zero string")
+    content_dir = Path(app.srcdir) / app.config.jupyterlite_content_dir
+    shutil.rmtree(content_dir, ignore_errors=True)
+        
     # Create the content dir
-    os.makedirs(os.path.join(app.srcdir, CONTENT_DIR), exist_ok=True)
+    content_dir.mkdir(exist_ok=True, parents=True)
 
     if (
         config.jupyterlite_bind_ipynb_suffix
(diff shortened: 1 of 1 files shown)


keewis:
reading the configuration docs, what `contents` refers to is quite confusing... on `main` there's `jupyterlite_contents` and `jupyterlite_ignore_contents`, as well as the hard-coded `CONTENT_DIR` that possibly is only used for `try_examples`? If so, I might rename the setting to `try_examples_content_dir` (or `jupyterlite_try_examples_content_dir`, not sure).

agriyakhetarpal (commented):
Hi @keewis, sorry about the delay here; it's been conference season for me recently. And I've never gotten around to playing with your xarray PR (as unfortunate as that is, from my end, apologies) :D

My memory also eludes me, but if I remember correctly, the intermediary folder where the notebooks are stored was named `contents` – which should be something like `jupyterlite_contents` indeed.

> clean up _before the next run instead of after the current run_

This makes sense to me. We could also escape a bit of this mess and make the intermediary folder a temporary directory per build...

keewis:
no worries, I've found a way to work around any delays: upload a conda package to a custom channel on prefix.dev, then use that until the PR has been resolved.

> the intermediary folder where the notebooks are stored was named `contents`

it was `_contents`, which was by default adjacent to `_build`.

> We could also escape a bit of this mess and make the intermediary folder a temporary

potentially, but I think it is nice to be able to check the generated notebooks in case anything went wrong. Another option would be to put it by default into `_build/contents` (or at least, below `_build`).

agriyakhetarpal (approved):
Thank you! Yes, let's not touch that any further; I agree with your reasoning to be able to check the notebooks.

I had no clue about uploading conda packages to prefix.dev channels. Looks like this is so much more easier now. Thanks for sharing that tip :)

martinRenou (approved):
Thanks!

